PR #51, opened 2024-08-19
Graph functions 2 testing

Amasus:
This pull request is mostly about a refactoring of graph_functions.r  (now graph_functions_deprecated.r) so that it reads data from the config files. Key changes include:
1. Automated descriptor generation
   1. This requires checking that every config folder varies in exactly 1 field 
   2. This was not previously true. Therefore,
      1. Config files have been rearranged
      2. A new config field has been added
  3. The automatically generated descriptor is informative but does not necessarily give descriptors that are useful for graphing. Thereofore in many of the graph description files, a second "change descriptor" step is added. This is a manual process, and feels like it should be up near the constants. **REVIEWER ASK:** Please review this process and give feedback
  4. **REVIEWER ASK:** If you find places where graph functions or map_functions is still identifying what to do based on grepl calls, please tag and ask me to fix. This is no longer an acceptable solution.
  5. **REVIEWER ASK:** Currently, the flow feels like the config files are being read in multiple times. While this doesn't affect the functionality, it seems inefficient. Thoughts welcomed.
1. Errors found in config files for Madison. The straight line versions of these have been corrected and rerun. The driving distances have not been.
2. Automated global driving distance generation. 
3. A new function to compare the placements across two different config folders (needed for Gwinnett_12_for_2024)
4. Refactoring of the 2 Madison and Savannah files to accommodate the above changes. Same for Gwinnett_12_for_2024. FFA_maps and FFA_graphs should now be deprecated, and replaced by basic_analysis
  1. I've made an attempt to combine the mapping and the graphing into one file in the above cases.
3. CLC_analysis and Engage_VA_analysis have not been refactored. They should be clearly marked as deprecated.
4. **REVIEWER ASK:** Before you approve these changes, please delete the graph_functions_deprecated file to keep the working tree clean.


antisocialscientist:
get_driving_flag and set_global_driving_flag need to be cleaned up

antisocialscientist:
rename intermediate config_file_name

antisocialscientist:
graph_functions.R line 235: update documentation

antisocialscientist:
Confirm before merging: The function combine_results has not changed, we are just making the final passed parameter a bit clearer to the reader?

Amasus:
> Confirm before merging: The function combine_results has not changed, we are just making the final passed parameter a bit clearer to the reader?

No, combine results is one of the files that has changed the most. 
* The if loop with "analysis_type" is no longer a thing. 
* Instead this function uses assign_descriptor to pull the appropriate descriptor from the the config data. 
   * There may be a better way to do the assign_descriptor. I'm not thrilled about needing to do a string match here. I'd love input into how this can/ should be managed.
* I've also since changed the order in which the list of dfs were being combined and the types were being changed, as a previous string search error (that shouldn't occur here, but again, string matches are making me nervous) was causing more files to be pulled than intended, and thus checking that a column existed in _just_ the first element of the list is inappropriate.

antisocialscientist (approved):
Thanks for the documentation fixes/explanations. This looks good to me now.

PR #54, opened 2024-09-17
Feature/modular distance calculations

orthorhombic:
## Note: This is a draft

This adds the capability (and a framework) to implement a number of new distance metrics.  Distances can be calculated either directly (point-to-point) or using isochrones to generate distances as a function of time.

Recognizing there is a lot of work happening on the config files at the moment, this should likely be merged (and refactored slightly) after those have been completed. hopefully this may inform some of that work. Eventually, I envision expansion of the distance data sources to include the the following:

```python
DATA_SOURCES = {
    "direct_distance": [
        "OrsDirectDistance",
        "HereDirectDistance",
        "HaversineDistance",
        "OsmDirectDistance",
        "GoogleMapsDirectDistance",
    ],
    "direct_time": [
        "OrsDirectTime",
        "HereDirectTime",
        "OsmDirectTime",
        "GoogleMapsDirectTime",
    ],
    "isochrone": [
        "OsmIsochrone",
        "HereIsochrone",
        "OrsIsochrone",
    ],
}
```

Additional documentation is forthcoming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,8 @@ A FEW THINGS TO NOTE:
 1. Create and activate conda environment. (Note, on a Windows machine, this requires using Anaconda Prompt.)
     1. `$ conda env create -f environment.yml`
     1. `$ conda activate equitable-polls`
-
+1. If updating the conda environment:
+    1. `conda env update -f environment.yml --prune`
 
 # To run
 From command line:
@@ -115,8 +116,8 @@ There are six files needed to run this program. The current Repo contains these 
 ### **Census Data (demographics and shapefiles)**:
 The sofware requires a free census API key to run new counties. You can [apply on the cenus site](https://api.census.gov/data/key_signup.html) and be approved in seconds. 
 
-    1. Create the directory authentication_documents/ 
-    2. Inside authentication_documents/ create a file called census_key.py
+    1. Create the directory authentication_files/ 
+    2. Inside authentication_files/ create a file called census_key.py
     3. The file should have a single line reading: census_key = "YOUR_KEY_VALUE"
 
 If you are only running counties already in the repo you can use the empty string for your key (census_key = "") but the censu_key.py file must still exist locally.
--- a/common/distance.py
+++ b/common/distance.py
@@ -1,0 +1,381 @@
+import pandas as pd
+import geopandas as gpd
+import time
+from epl_io.osm import OsmIsochroneGenerator, CoverageError, OsmDistanceCalculator
+from epl_io.haversine import HaversineDistanceCalculator
+from typing import Union, List
+from model_config import OsmArgs, ConfigOtherArgs
+
+DATA_SOURCES = {
+    "direct_distance": [
+        "OrsDirectDistance",
+        "HereDirectDistance",
+        "HaversineDistance",
+        "OsmDirectDistance",
+        "GoogleMapsDirectDistance",
+    ],
+    "direct_time": [
+        "OrsDirectTime",
+        "HereDirectTime",
+        "OsmDirectTime",
+        "GoogleMapsDirectTime",
+    ],
+    "isochrone": [
+        "OsmIsochrone",
+        "HereIsochrone",
+        "OrsIsochrone",
+    ],
+}
+
+
+class BaseDistanceGenerator:
+    OUTPUT_COLUMNS = [
+        "id_orig",
+        "id_dest",
+        "distance",
+        "county",
+        "address",
+        "dest_lat",
+        "dest_lon",
+        "orig_lat",
+        "orig_lon",
+        "location_type",
+        "dest_type",
+        "population",
+        "hispanic",
+        "non-hispanic",
+        "white",
+        "black",
+        "native",
+        "asian",
+        "pacific_islander",
+        "other",
+        "multiple_races",
+        "weighted_dist",
+    ]
+
+    def __init__(
+        self,
+        state_name,
+        county_name,
+        data_source,
+        origins,
+        destinations,
+        other_args: ConfigOtherArgs,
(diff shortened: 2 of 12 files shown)


Amasus:
Hey @orthorhombic, 

Sorry for not getting to this. I just noticed that you'd made a PR, but since you hadn't tagged me to review, I hadn't received an email notification.

Below, I'm including some initial thoughts about the state of the code. If you aren't ready for this yet, then please, by all means ignore what is to follow. I hate it when people comment of work that hasn't been finished yet, and am aware that I might be doing the same to to. OTOH, I don't want to let this languish when you are expecting a response. Feel free to tell me how you wnat to proceed.

* I see what you are trying to do with other args, and I think I agree with you in principal. However, in Ben's work, the other_args are being used as a staging area for config fields that are still in prototype. Eventually, one hopes that having the various distance sources will graduate from other_args to the main database.
* I like what you've done with the epl_io. Am I understanding correctly that any other distance function I want to put in (log_haversine, log_driving, etc.) should go in as a class like this?
   *  This is the fundamental architechture decision that I'd love to get from you. `What is the correct, pythonic way to modularize different distance metrics?
   * The main distances that I'm interested in at the moment is driving, haversine, log_driving and log_haversine. 
   * Is the right way to do this to add more cases to the class `DirectDistanceGenerator`?

I'd love to talk to you more about this, and figure out how to get this incorporated into main. As you noted, before proceeding, we'll need to figure out what is going on with the config files first (which is still in flux, though settling down). 

Let me know how you want to proceed.

orthorhombic:
Thank you for the feedback! Below I've addressed some of your thoughts. As a next step, I think it is worth discussing how (and when) we promote new config arguments out of being "prototype" as well as get some additional thoughts on the modular approach I've implemented here. I'd be happy to jump on a call to discuss with anyone interested.  I also plan on updating this with proper docstrings prior to merging.


 > I see what you are trying to do with other args, and I think I agree with you in principal. However, in Ben's work, the other_args are being used as a staging area for config fields that are still in prototype. Eventually, one hopes that having the various distance sources will graduate from other_args to the main database.

I agree we'd want to identify an approach for these config arguments to graduate from other_args. As a caveat to the following thoughts, I'd like to see the database migration complete before we settle on an exact process since I don't want to cause any delays (or added burden) there. As we eventually move to support additional metrics, I think we will likely need to keep a structured json object that fully defines the parameters we choose for distances and leave the validation of such to the python code. This way each data_source can have different parameters (and expand) as we implement features or APIs change. As examples, we may eventually want to parameterize the distance methods with transportation method, time of day, "snap to road". These are going to vary from one source to the next. 

To move distance arguments out of `other_args`, the most simple change we could make here is to just change `other_args` as I use it here to something like `distance_args` and promote that as a single key. I think we have flexibility here. The exact save/load method for a json object like that in the new database might need to be iterated on.


> I like what you've done with the epl_io. Am I understanding correctly that any other distance function I want to put in (log_haversine, log_driving, etc.) should go in as a class like this? 

I would agree any new method of calculating distances should be implemented as a class we plug into this framework. Specifically I envisioned a few different types of component:
- Distance Calculator: This is a class where we initialize a data source (and travel parameters). Once initialized, this can use standard methods to calculate the distance between two locations.  
- Isochrone Generator: This is a class where we initialize a data source and can generate isochrone shapefiles for specific locations (and travel parameters).
- Distance Generator: The purpose of this is to take a standard type of Distance Calculator or Isochrone Generator and apply that calculator to everything in the supplied dataset and give the resulting distance outputs in a standard format. This fundamentally assumes a batch process. Currently I can only think of two types, but maybe we'll think of more. "Direct" generators use a DistanceCalculator which calculate an "A-to-B" type distance. "Isochrone" generators specifically use shapefiles (from an IsochroneGenerator) to calculate distances through a many-to-one iterative process.


> This is the fundamental architechture decision that I'd love to get from you. `What is the correct, pythonic way to modularize different distance metrics?

There is a saying: "The best camera is the one you have with you." Applying the spirit of that here, I think that the best way to modularize the metrics is the one that makes sense to people and gets used. To that end, this makes sense to me where we define a BaseDistanceGenerator which specifies the outputs and the calc method which needs to be implemented by the specific type of generator. I think we could probably find a more optimized approach and get fancy with decorators and more complicated inheritance, but I'm not convinced the complexity is worth it. I'd also speculate we'll  refactor this some as we grow into implementing other distance sources. In my mind, the most important part is that we move to modular components where we have well defined inputs/outputs and standard methods. That will help us grow and make things easier to test.

This is totally ignoring potential speed optimizations we might need to address, but as of now, I don't think this is a big pain point. I also plan on reading a little more about design patterns to see if any alternatives jump out at me.

> The main distances that I'm interested in at the moment is driving, haversine, log_driving and log_haversine.

That makes sense. Right now, haversine is implemented in my code. In my thinking, the `haversine` calculator would be the same distance calculator as `log_haversine` since I'm guessing it's just a simple transformation. Rather than saying they are two different data sources, I think it might make more sense to add a `log` option in one of the generator classes (probably BaseDistanceGenerator). Then we could pass the log option and get that for every distance/time generated (same with any other transformation). When you say `driving`, I think that would fall under one of the `direct_distance` or `direct_time` data sources that I proposed (or exist as a new value under those keys). This way, by specifying the source, we can validate that the source has the right source-specific parameters. I think it also would reduce some confusion between different driving distance data sources (OSM, ORS, HERE, Google, etc).

> Is the right way to do this to add more cases to the class DirectDistanceGenerator?

For any new source, you are correct it would be a new case in DirectDistanceGenerator. It would also need a new Distance Calculator class to handle the nuances of how that specific distance gets calculated (or sent off to an API).

PR #56, opened 2024-10-20
Osm google maps compare

jmlar:
The GMAP_ README documents the code and key points.  It also provides setup instructions for local OSM servers


Amasus:
@jmlar, a few thoughts from reading the readme:
1. I can't find the file `analyze_driving_distances.R`. Am I missing something?
2. Can the files be refactored so that, instead of putting the API key in `authentication_files` , it it put in `authentication_files/GMAP_Platform_KEY.py` with content `GMAP_Platform_KEY=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx`? This would keep it consistent with the census key. If you have a different way you'd want to do this, I'm all ears. Consistency is more important than the specific method.
3. So far, I'm seeing only the compare_driving_distance in Virginia_Beach. Is that correct?

I'll take a look at other files next, as I go through this piece meal, but I wanted to give you incremental requests. Happy to discuss anything async (discord/ email is the best). Feel free to push back on anything you don't like here

antisocialscientist:
One more quick thought on the readme. Do you have specific versions of the required packages so we can fold them into the equitable-polls virtual environment?

Amasus:
Dear @jmlar, 

I've reviewed GMAP_data, and it looks good and makes sense to me. There's one feature i'd like for you to add, if possible. 

When you create sample and sample_summary, you pull data that is in the correct quartile for the variable you are interested int, which is great. But, I believe that the 500 rows saved, say for the lowest population density quartile only has data for population density, and not for pct white and income as well. Is there a way to merge the data so that every row in the final data set has data for all three demographic features? This will make future regression analysis eaiser on this work. 

Other than that, this all looks great. Thank you for all your work.

--Susama

jmlar:
@Amasus  
1) I renamed analyze_driving_distances.R. to GMAP_distance_report.qmd but forgot to update the documentation.  The  README has been corrected.   The authentication_file has also been refactored.  You'll see the change withthe next push.
2) Just to confirm, it sounds like you’re asking that each row in the sample dataset contain values for all demographic features—population density, pct white, and income—from the original data. The current subset already includes all this information, so I believe it has everything needed for your regression analysis. Or am I missing something?

@antisocialscientist which software are you thinking about?  I assume python, but are you also thinking about R and/or OSM?

Amasus:
@jmlar, 
okay, i see now that you are reading in the county_ST file, which is great. But this raises an issue I'd missed earlier, and maybe answers the original question

New issue:
We'd wanted to do this work with the lowest quartile of population density, not lowest quartile population... The block area is saved in the files in datasets/census/tiger/county_ST/, and I'm not seeing where that is being called at the moment.

Clarification of the old issue:
While I agree that the County_ST.csv files have the population, and racial demographic areas, it does not have the economic or population density data in them. I see where you are pulling joining the economic data to the County_ST.csv (lin 312-319), and then using this to sample. Therefore the file outputted should have all the information in it, along with the difference in driving distances, correct? 

If so, can you just change the population to population density?

jmlar:
@Amasus 

Got it.  Will modify code to use Tiger files to extract landmass and convert to population density.

jmlar:
I just pushed the code with population density added and other code review changes made.

jmlar:
Added an option to recalculate OSM driving distances

PR #60, opened 2024-11-10
Feature/analysis from db

Amasus:
Okay, take 2 at making a pull request for refactoring the R code to read data from the data base in order perform the analysis.

@antisocialscientist , this was the set of files you were supposed to have reviewed a couple weeks ago while flying to Mexico. Please let me know if there are issues. This should be a direct translation of the work you did for the config file review, with the exception that now, instead of calling things from local directories multiple times, data is called once from the remote DB and written to local memory to avoid expense.

@abd1tus I'm putting you on this for context only, so you have an idea of how I was hoping to use the DB you are designing to do analysis... Things here can be changed later if your vision differs enough from what Benjamin was doing to need a refactor.


Amasus:
@antisocialscientist you are currently blocking @abd1tus . If you cannot get to this, please let me know and I'll find another solution

PR #62, opened 2024-11-11
For paper/dekalb driving

Amasus:
This PR does two things:
1.  merging the work that @dskipper did for the ESRA paper into the format that is needed for the database migration
2. changes the logic of what rows are labeled polling and potential to be able to avoid the manual relabeling work that Daphne had to do to run the files.

I'd like @dskipper and @abd1tus to review differet parts of this PR.

1. @abd1tus, There are a bunch of DeKalb config files and associated results that are being brought into the set of things that need to be updated here. Please check that I've renamed things correctly, and migrate these configs/ results as well
2. @abd1tus , @dskipper in model_data.py, in the clean_data function, i've changed how the labels "polling" and "potential" are assigned to locations. 
   1. Specifically, it is no longer read in form the file. We may eventually want to change how we defined the *_locations_only.csv files to reflect this
   2. Instead, the dest_type column is set to "polling" if and only if the year we are interested in is in the "location_type_column". Otherwise, it is set to "potential", if it is a building or "bg_centroid" if it is a centroid
   3. This should allow you to select 2020 and 2022 locations for the model, and treat any 2024 locations a potentials, which is the desired behavior


dskipper:
Hi Susama,

I can't see in this code from clean_data() where the polling locations that are not in the year_list are being added as potential locations. I could be missing that, though, or maybe it is happening elsewhere?

Also I didn't see this in the file diff, I had to go to the model_data.py  file directly to read this code, so I'm not completely sure I'm seeing the correct code.

Screenshot 2024-11-11 at 10 35 21 AM


--
Daphne Skipper, Ph.D. 
Associate Professor
Mathematics Department
Senior Academic Adviser, Operations Research
United States Naval Academy
(pronouns: she/her)

Amasus:
Not sure why you are seeing what you have posted. Have you pulled the
branch from remote to local?

here's what I'm seeing when i look at the compare code on git:
[image: image.png]

dskipper:
OK, I wasn't in the correct branch, but I just looked and that portion of clean_data() looks good to me. I don't know why the compare code feature wasn't looking right to me? I'm sure I was doing something wrong but don't really have time to troubleshoot.

PR #64, opened 2024-12-12
single config config_sets now run

Amasus:
@antisocialscientist Can you take a quick look at this?

Issue to be addressed:
The way that basis_analysis.r was processing files, it did not have a good way to handle config_sets (config folders) with only one config in it. Namely, if there are multiple configs, it automatically figures out what the field of interest is from the field that varies. But with 1 config, it can't figure this out

Solution:
basis_analysis.r
Extra constants (indicating Field of interest) have been added to the front of basic_analysis.r
the calls to read_result_data now take these contants explicitly
in graph_functions.r
process_configs_dt has new logic build in to check that the constant is both provided when expected, and is a valid field


antisocialscientist (approved):
This looks straightforward. The fixing the for loop and turning it into applys is nice,

PR #65, opened 2024-12-12
Feature/make maps every time

Amasus:
Issue to be addressed:
1. When we move to the database, the folder map_work will not be migrated, particularly since this was full on intermediate steps
1. Upon further reflection, it occured to me that, since we aren't really using cartograms any more... we don't need to load any map data. This was primarily there to keep from having to make a cartogram every time.

Solution:
In map_functions.R:
make_or_load_maps no longer checks whether a file exists. It just creates the maps. 
all references to cartogram and the folder_map_work has been removed.


antisocialscientist (approved):
Clean and clear. I am a bit sad to see the cartogram functionality go, but as it wasn't useful to the actual clients no reason to keep trying to support it because I like it.

PR #71, opened 2025-03-15
Feature/code split out

abd1tus:
Directory structure cleanup.  All of the python code has been put under a folder "python" and all the R code under the "R" folder.


Amasus:
Comments:

- There is a results_analysis, a results_analysis_output and an R/results_analysis folder. Only the latter 2 should remain in the repo moving forward. Nothing should write to results_analysis.
- What is `python/database/imports.py` doing. It seems to be a completely new file.
- Same for `python/database/query.py`
- have you tested the address validator? I have not
- why is import_edes now deleted in `python/scripts/db_import_cli.py`?
- is the commented out stuff in `python/scripts/db_import_locations_cli.py` or is that just old code you haven't removed yet?
- FULL_DF_COLS (line 214 of model_data.py) is also a set of constants that should be pulled out
- I'm not worrying about things in the tests folder because
   - you said that tests are failing right now
   - Getting testing right should be its own effor
   - this effort should include getting some sort of CI pipeline set up to avoid this problem in the future
- `RESULTS_BASE_DIR = os.path.join(PROJECT_ROOT_DIR, 'datasets')`
   -  this is a copy paste error where you've set RESULTS_BASE_DIR to be the same as DATASETS_BASE_DIR. 
      - It should actually be something like `os.path.join(PROJECT_ROOT_DIR, 'datasets', 'results')` assuming that we want the datasets folder to contain both inputs and outputs. 
      - Otherwise, we'd need to make another `os.path.join(PROJECT_ROOT_DIR, 'model_outputs') folder or similar.
   - Do we need to create this directory for the code to run in -CSV mode? I have not yet tested this
- `python/utils/pull_census_data.py` has the note that you need to do something for the authetication_key. Has this been done? I don't see it.

abd1tus:
> What is python/database/imports.py doing.  It seems to be a completely new file.
> Same for python/database/query.py

Originally there was a db.py which contained most the queries as well as some code from imports.  With the python directory split out there were errors arising from circular dependancies.  imports.py was created to consolidate anything involving imports and separate it out from db.py.  db.py got renamed to query.py to avoid from python.database import database (which just seemed weird).

> have you tested the address validator? I have not

No, I have not.

> why is import_edes now deleted in python/scripts/db_import_cli.py

That code got moved to python/database/imports.py to avoid circular dependancies.  import_edes was being used in model_results.py and it really didn't make sense to keep those methods defined it in db_import_cli.py.  So general cleanup, but mostly to avoid circular dependancies.

> is the commented out stuff in python/scripts/db_import_locations_cli.py or is that just old code you haven't removed yet?

This should not have been added to the commit, it's new code for feature/import_data_sources that wasn't committed there so it didn't get stashed away.  I will delete it from this branch.

> FULL_DF_COLS (line 214 of model_data.py) is also a set of constants that should be pulled out

Eventually, yeah.  I wanted to avoid things that weren't related directly to the split out or avoiding circular dependencies. 

> RESULTS_BASE_DIR = os.path.join(PROJECT_ROOT_DIR, 'datasets')

I'll fix this.

> python/utils/pull_census_data.py has the note that you need to do something for the authetication_key. Has this been done? I don't see it.

I wanted to pull this out from being hard coded in the new structure, but I avoid more changes.  But it makes sense to clean this up here.  I'll work on it.

Amasus (approved):
Looks good.

PR #76, opened 2025-05-01
Feature/refactor graphs maps

Amasus:
Broke graphs functions out into 
* load_config_data
* graph_functions
* regression_functions

Significantly redid how map_functions work.
Of note:
* We do not read data in differently for maps vs boudaries.
* The old make_or_load_data is still there, now make_or_load_data_old
* Please check the difference in logic that confirm that you don't think I've done something stupid

The goal was to document (for graph functions) 
and to streamline (for map functions).

Better ideas for documentation / organization and streamlining are actively desired.

### DIFF
--- a/R/result_analysis/Basic_analysis.r
+++ b/R/result_analysis/Basic_analysis.r
@@ -110,7 +110,7 @@ global_color_bounds <- distance_bounds(all_prepped_output)
 orig_regression_data <- get_density_data(orig_output_df_list$result)
 potential_regression_data <- get_density_data(potential_output_df_list$result)
 
-#take density data and aggregate key columns up to the block level
+#take density data and aggregate key columns up to the block group level
 orig_bg_density_demo<- lapply(orig_list_prepped, function(df)bg_data(df))
 potential_bg_density_demo <- lapply(potential_list_prepped, function(df)bg_data(df))
 
--- a/R/result_analysis/Basic_analysis.r
+++ b/R/result_analysis/Basic_analysis.r
@@ -21,6 +21,7 @@ setwd(here())
 source('R/result_analysis/utility_functions/storage.R')
 source('R/result_analysis/utility_functions/graph_functions.R')
 source('R/result_analysis/utility_functions/map_functions.R')
+source('R/result_analysis/utility_functions/regression_functions.R')
 
 
 #######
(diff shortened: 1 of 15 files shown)


antisocialscientist (commented):
Basic_analysis
- 38: I would change the error message to "must be exactly one config file"
- 96-98 I assume this is in order to have a range that covers highest possible values. This should either be confirmed in a comment or the logic could be clarified
- I'm wondering whether it is possible to aggregate the data up to block groups once instead of twice.

extract_city
- Does the user have to keep anything in mind for non-convex cities or those with large bodies of water here?
- I wonder whether it is worth it at some point to make this (and other later places) able to read the one-header as well as two-header census tables
- 21-28 Also, will you at some point want to pull the hand coded constants into the configs? Probably not now, but as a future change?

graph_functions
- 14 I really like the DEMO_COLS as a constant list
-  43-52 Just confirming that you want FALSE for whenever null_arg is not null, regardless of HISTORICAL_FLAG. Assuming so, I think that it might be clearer to reverse the logic order, first checking historic flag and then null_arg without ever having an AND
- 128&132 Maybe change the wording between the two  so that the error message is easier to read
- 219 & 257 associated (instead of assocaited)
- 241 Would it be overkill to explicitly include the by in the merge? I am imagining a situation where you manage to accidentally merge things with an extra column name in common and then the outputs are inexplicably weird

load_config_data

- 113 I am assuming PRINT_SQL is a debugging flag that shows up just here for now but could show up elsewhere later

map_functions
- 193 any input input NULL?

Amasus:
Basic_analysis

- [x ] 38: I would change the error message to "must be exactly one config file"
- [x ] 96-98 I assume this is in order to have a range that covers highest possible values. This should either be confirmed in a comment or the logic could be clarified
- [ ]I'm wondering whether it is possible to aggregate the data up to block groups once instead of twice.

extract_city
Not part of this MR

graph_functions
- [x ] 14 I really like the DEMO_COLS as a constant list
- [ x] 43-52 Just confirming that you want FALSE for whenever null_arg is not null, regardless of HISTORICAL_FLAG. Assuming so, I think that it might be clearer to reverse the logic order, first checking historic flag and then null_arg without ever having an AND
- [x ] 128&132 Maybe change the wording between the two so that the error message is easier to read
- [x ] 219 & 257 associated (instead of assocaited)
- [ x] 241 Would it be overkill to explicitly include the by in the merge? I am imagining a situation where you manage to accidentally merge things with an extra column name in common and then the outputs are inexplicably weird 

load_config_data

- [x]113 I am assuming PRINT_SQL is a debugging flag that shows up just here for now but could show up elsewhere later
map_functions
   - PRINT_SQL is defined in the config file
- [x]193 any input input NULL?

antisocialscientist:
The two places the aggregating to block groups is done is inside prepare_outputs_for_maps on line 94/5 and then again more directly for regressions on 114/5. (Also the comments on both omit group from block group.)

antisocialscientist (approved):

PR #80, opened 2025-06-21
SA attempt to cleanup main

Amasus:
Got rid of a few files that shouldn't be sitting out in main. Have the following questions:

- There are multiple .gitignores.
   - one in equitable-polling-locations
   - one in equitable-polling-locations/python
   - is this desired?
- similarly, there are multiple 
- in equitable-polling-locations/.gitignore, we have the following:
   - `#Want to avoid keeping config files locally
      #and put them in the database
      *.yaml` (lines 188-190)
   - but we still have files in equitable-polling-locations/datasets/configs
   - What do we actually want here?
     - to have these config files to only be local?
     - to have only the template config (for reference) in main, and the rest local?
        - this should be easy as this isn't a .yaml file
     - to only have the *.yaml_template files, but nothing else?
  - note, currently the python/tests directory has its own .yaml files, which I don't _think_ we want in the database?
- Do you know what `datasets/census/tiger /Untitle.iypnb` is? If not, can we get rid of it?
- `datasets/polling/` still has .csvs in it. is this desired?
   - line 181 of `equitable-polling-locations/.gitignore` has .csv commented out. This was because the migration for input files hadn't fully been moved. Should we uncomment this now?


abd1tus:
**Regarding the .gitignores**

The .gitignores can be consolidated.  Originally I moved the one from project root to python so it could focus on python specific.  But between merges it looks like the one in the project root got re-added.  I'll clean the two up and consolidate.

**Regarding the configs**
I put together a list of configs in main that haven't been imported into the db.  I'll work on getting them into the prod db. Once done we should remove the configs from the repo, but we should keep the templates since there is no current db import from templates.  

The test config yaml files should remain in the repo, correct.

**Untitle.iypnb**
We can get rid of it, likely a scratch left over when from when we using google colab. I'll delete it.

**datasets/polling**
We can clear this out since we only need the locations_only csv files and they are all imported into the db.  Yes, we should put .csv back in gitignore.  I'll clear them.

abd1tus:
Changes in.  .gitignores are cleaned up, configs removed, removed the Untitled.ipynb, and cleared locations.

PR #130, opened 2025-07-27
Feature/test fixes

abd1tus:
Significant test cleanup, fixes, and additions.  Converted incorporate_penalties to a class instead of one long method to make testing easier.


python/scripts/db_import_driving_distances_cli.py
Amasus, line 1: Probably not for now, but eventually, I'd love to get rid of the need for this

python/tests/model_data_test.py
```diff
@@ -0,0 +1,189 @@
...
+        if not dist_loc == pytest.approx(dist_drive):
+            mismatches.append(
+                f'  (id_orig=\'{orig}\', id_dest=\'{dest}\'): '
+                f'locations_results_df={dist_loc}, driving_distances_df={dist_drive}'
+            )
+
+    assert not mismatches, \
+        'Distance mismatches found for the following (id_orig, id_dest) pairs:\n' + '\n'.join(mismatches)
+
+def test_build_source_column_output(driving_locations_results_df):
```
Amasus, line 106: Would it make more sense for this to be done in R and then compared like we do in other cases?

python/tests/model_run_test.py
```diff
@@ -1,116 +1,178 @@
...
-    assert len(EX_OPEN_PRECINCTS - set(PENALTY_CONFIG.penalized_sites))==3
-
-def test_penalized_model():
-    penalty = (EX_KP - KP) / len(OPEN_PRECINCTS)
-    PEN_MODEL = model_factory.polling_model_factory(DIST_DF, ALPHA, PENALTY_CONFIG,
+    # pylint: disable-next=line-too-long
+    assert all(dest_pop_df.population <=(polling_locations_config.capacity*total_population/polling_locations_config.precincts_open))
+
+
+# # Test the intermediate dataframe with driving distances
```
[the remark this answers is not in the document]
abd1tus: Will remove.

Amasus (approved):

PR #150, opened 2025-09-14
modified build source details, checked max_min_mult and capacity

Amasus:
- Config templates for testing max_min_mult:
    - datasets\configs\Gwinnett_County_GA_max_min_mult_test
- Config templates for testing capacity:
    - datasets\configs\Gwinnett_County_GA_capacity_test
- Modification to db_import_cli error message
- build_source creates the appropriate folder if it doesn't exist
- model_run now writes results to the correct path

### DIFF
--- a/python/solver/model_data.py
+++ b/python/solver/model_data.py
@@ -383,7 +383,7 @@ def build_source(
     full_df['id_dest'] = full_df['id_dest'].astype(str)
     
     if not os.path.exists(output_path):
-        os.makedirs(os.path.join(POLLING_DIR, location))
+        os.makedirs(os.path.basename(output_path))
     full_df.to_csv(output_path, index = True)
     return result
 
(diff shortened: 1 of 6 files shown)


python/solver/model_data.py
```diff
@@ -376,12 +376,14 @@ def build_source(
...
     #####
-    #reformat and write to file
+    #reformat and write to file (making directory if it doesn't exist)
     #####
 
     full_df['id_orig'] = full_df['id_orig'].astype(str)
     full_df['id_dest'] = full_df['id_dest'].astype(str)
     
+    if not os.path.exists(output_path):
+        os.makedirs(os.path.join(POLLING_DIR, location))
```
abd1tus, line 386: This would be better if it was:

`os.makedirs(os.path.basename(output_path))`

 since output_path is not guaranteed to use os.path.join(POLLING_DIR, location) if output_path_override is used.

abd1tus (approved):

PR #165, opened 2025-12-17
regressions now save and upload to DB

Amasus:


Amasus:
Superseded by #346 — same underlying idea (extracting closure-regression plotting out of the exploration script), rebuilt against the 2026 data with the fixes this branch needed (schema drift, title wrapping, non-interactive GCS auth).

Issue #32, opened 2024-03-20
Use location names in Google queries for lat lon


abd1tus:
Update script to query google with location names for better accuracy.

Issue #31, opened 2024-03-16
Why do some runs not match all residences?


Amasus:
Sometimes the optimization run will not properly complete, and return a value anyway. 

Specifically, it will match some (but not all) of the residences to precincts. E.g.:
1. DeKalb_GA_no_school_configs\DeKalb_config_no_school_27.yaml
2. Greenville_SC_original_configs\Greenville_config_original_2016.yaml

Rerunning the .yaml file doesn't seem to fix the problem, so this _should_ be reproducible. 
To see that all the blocks aren't matched, go to the appropriate residence_distance.csv and note that the number of unique residences in that file is different from the number of unique residences in other files associated to the same county.

I cannot figure out why this is happening. The constraints as defined should not allow this to occur. Specifically, res_assigned_rule in model_factory.py is the constraint that specifically insists that each block is assigned to exactly one precinct.

Amasus:
Resolved. See commit with [hash 416f391](https://github.com/Voting-Rights-Code/Equitable-Polling-Locations/commit/416f391455ddc14655737b088ab4b272ffe4ce36)

need to merge commit with hash 416f391 to main (and elsewhere, as this is what is causing the bug of some blocks being dropped)

DO NOT CLOSE BEFORE DOING THE CHERRYPICK PROPERLY.

PR #200, opened 2026-04-03
feat(tests): add end-to-end CLI test suite

abd1tus:
## Summary

- Add comprehensive end-to-end test suite for all CLI scripts (`model_run_cli`, `model_run_db_cli`, `auto_generate_config`, `db_import_*`) with session-isolated test data, shared DB fixtures, and value-level assertions
- Add `python run.py e2e_tests` command for Docker-based e2e test execution with `e2e_csv` and `e2e_db` pytest markers
- Fix `model_run_db_cli` environment loading bug and distance data import to use pre-computed CSVs instead of rebuilding from Census shapefiles
- Add operational rules, child CLAUDE.md files for Python and R conventions, and Claude Code settings

## Changes

- **`python/tests/e2e/`** — New e2e test package with `conftest.py` (session fixtures, `run_cli()` helper, isolated CSV/YAML generation, DB import fixtures) and test modules for each CLI script
- **`run.py`** — `e2e_tests` command that forwards pytest markers via `-m` flag
- **`pytest.ini`** — Register `e2e`, `e2e_csv`, `e2e_db` markers
- **`python/scripts/model_run_db_cli.py`** — Fix: always load environment since DB configs require it regardless of output type
- **`CLAUDE.md`, `python/CLAUDE.md`, `R/CLAUDE.md`** — Restructure docs into root + child convention files
- **`.claude/settings.json`** — Sandbox config, permissions, network allowlist
- **`docs/`** — Design spec, implementation plan, testing docs

## Test plan

- [x] `python run.py e2e_tests -m e2e_csv` — all CSV-based e2e tests pass without DB credentials
- [ ] `python run.py e2e_tests -m e2e_db` — all DB-based e2e tests pass with GCP credentials configured
- [ ] `docker compose run --rm app pytest` — existing unit tests unaffected
- [ ] `pylint python/` — no new warnings


docs/to_install.md
```diff
@@ -24,9 +24,75 @@ docker compose run --rm app pytest
 All tests should pass.
 
 
+## Testing
```
Amasus: This section being here makes the installation instructions complicated.

Recommendation:
1. Add to the existing instructions on how to run the necessary tests to check that installation occurred correctly.
2. Put the testing information in a different file. Say, to_test.md. 
    1. This file should have more context about how different tests are organized etc. E.g. if there are reasons one would not want to run the e2e tests. etc. 
4. Link to that file for further instruction

python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py
```diff
@@ -24,8 +24,8 @@ def upgrade() -> None:
     # Note: RENAME COLUMN has to happen first, doing the other way around
     # causes the column rename to fail sliently
     op.execute(
-        f'ALTER TABLE `{db_dataset}.polling_locations_sets`'
-        f'RENAME COLUMN `polling_locations_only_set_id` TO `potential_locations_set_id`'
+        f'ALTER TABLE `{db_dataset}.polling_locations_sets` '
```
Amasus, line 27: I'm confused as to what is going on here.
* If I look at the GCP, I see a table `potential_locations_sets` with a column `potential_locations_set_id`, as I expect. 
* Unclear to me why we are tallking about polling_locations_sets. I don't see this dataset in the database at all. **is this deprecated?**
   * certainly, the `locations_only_set_id` is deprecated and should not be appearing here.
* Either way, I see that the documentation in the `docs/database.md` refers to a `polling_locations` dataset, and not a `potential_locations` dataset, so that needs to be updated.

python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py
```diff
@@ -41,7 +41,7 @@ def downgrade() -> None:
     # causes the column rename to fail sliently
     op.execute(
         f'ALTER TABLE `{db_dataset}.distance_data_sets` '
-        f'RENAME COLUMN `potential_locations_set_id` TO `polling_locations_only_set_id`'
+        f'RENAME COLUMN `potential_locations_set_id` TO `locations_only_set_id`'
```
Amasus, line 44: Again, confused. Looking at our GCP I see:
* `distance_data_sets` has a column `potential_locations_set_id` (as expected) and not a `polling_locations_set_id`
* In either case, the column name should NOT be `locations_only_set_id`. That is definitely deprecated.

python/database/alembic/versions/5d5706c75d62_rename_polling_locations_only_to_.py
```diff
@@ -24,7 +24,7 @@ def upgrade() -> None:
     # causes the column rename to fail sliently
     op.execute(
         f'ALTER TABLE `{db_dataset}.polling_locations_only` '
-        f'RENAME COLUMN `polling_locations_only_set_id` TO `potential_locations_set_id`'
+        f'RENAME COLUMN `locations_only_set_id` TO `potential_locations_set_id`'
```
Amasus, line 27: Again, same concerns as in `python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py`: 
* I'm not seeing `polling_locations_only` in the GCP, only `potential_locations_only` (as I expect)
* the column name should NOT be `locations_only_set_id`

python/database/alembic/versions/5d5706c75d62_rename_polling_locations_only_to_.py
```diff
@@ -40,7 +40,7 @@ def downgrade() -> None:
     # causes the column rename to fail sliently
     op.execute(
         f'ALTER TABLE `{db_dataset}.potential_locations` '
-        f'RENAME COLUMN `potential_locations_set_id` TO `polling_locations_only_set_id`'
+        f'RENAME COLUMN `potential_locations_set_id` TO `locations_only_set_id`'
```
Amasus, line 43: When I look at `potential_locations` in the GCP, I see the column `potential_locations`, not `polling_locations_only_set_id` (as expected)

* again, is `python/database/alembic/versions/5d5706c75d62_rename_polling_locations_only_to_.py` deprecated?
* either way, the column should NOT be `locations_only_set_id`

R/CLAUDE.md
```diff
@@ -0,0 +1,27 @@
...
+| `result_analysis/utility_functions/` | Shared functions: storage, graphs, maps, config loading |
+| `result_analysis/deprecated/` | Archived historical analyses |
+| `tests/` | Manual verification scripts (not automated) |
+
+Key libraries: `data.table`, `ggplot2`, `sf`, `bigrquery`, `googleCloudStorageR`, `yaml`, `plotly`. No formal R package structure — standalone analysis scripts.
+
+## Conventions
+
+- Follow the tidyverse style guide. Use `snake_case` for all names. Files: `snake_case.R`.
+- Assignment: always `<-`, never `=`. Use `TRUE`/`FALSE`, never `T`/`F`.
```
Amasus, line 19: never is too strong. There are time when, say, setting assignments in ggplot, one actually wants =. Is Claude smart enough to figure that out? 

I guess, for now, we can try it, and see what I have to change.

R/CLAUDE.md
```diff
@@ -0,0 +1,27 @@
...
+- All `library()` calls at the top of each script. No `setwd()` — use relative paths or `here::here()`.
```
Amasus, line 20: This is explicitly changing the convention that I've been using of 
```
library(here)
setwd(here())
```
Is that intentional? I'm fine if this is considered a better standard. Just want to be intensional.

CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+- **Warnings are errors** — fix all Pylint warnings before committing. Do not add `# pylint: disable` without a comment explaining why. Run pytest with `-W error`.
+- **Readability over brevity** — spell out variable names (`polling_location` not `pl`, `equity_score` not `eq_s`). Short names only for loop counters and lambdas.
+- **Conventional Commits** — format: `type(scope): description`. Types: `feat`, `fix`, `refactor`, `test`, `chore`, `docs`. Scope: module name (`solver`, `database`, `scripts`).
+- **Before merging** — run `python run.py test` and `pylint python/`. All tests and lints must pass.
+
+## Commands
+
+**Run the model (local files):**
+```bash
+python run.py model_run_cli -c NUM -l ./datasets/configs/<County>/config.yaml
```
Amasus, line 26: I think this path should be 
./datasets/configs/<config_set>/<config_name>.yaml

CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+python run.py model_run_db_cli -e ENV -c NUM -l config_set/config_name
```
Amasus, line 27: I don't think we need the -e ENV here.
I left a comment to the end in the features/readme_clarification as well

CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+See `docs/to_run.md` for full usage details.
+
+## Architecture
+
+```
+python/
+  solver/     # KP optimization core: config, data prep, penalties, solver, results
+  scripts/    # CLI entry points (model_run_cli, model_run_db_cli, db_import_*, etc.)
+  database/   # SQLAlchemy BigQuery ORM + Alembic migrations
+  utils/      # Env handling, GCP credentials, directory constants
```
Amasus, line 71: utils also is where the functions to pull population / demographics data (census, CVAP, ACS (future)) lives)

CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+```
+
+See `docs/to_run.md` for full usage details.
+
+## Architecture
+
+```
+python/
+  solver/     # KP optimization core: config, data prep, penalties, solver, results
+  scripts/    # CLI entry points (model_run_cli, model_run_db_cli, db_import_*, etc.)
```
Amasus, line 69: This is also where auto_generate_configs lives. I don't know if that counts as a cli entry point or not.

CLAUDE.md
```diff
@@ -0,0 +1,95 @@
...
+  database/   # SQLAlchemy BigQuery ORM + Alembic migrations
+  utils/      # Env handling, GCP credentials, directory constants
+  tests/      # pytest suite; fixtures in conftest.py
+    e2e/      # End-to-end CLI tests (subprocess-based, session-isolated)
+run.py        # Docker wrapper; auto-discovers scripts in python/scripts/
+datasets/     # Sample data by county: configs, polling locations, distances
```
Amasus: I wouldn't call this "sample data". Maybe:
```datasets/    #Where all the input, intermediate and output datasets for the alorithm lives. This includes where census datasets are stored, config files are written. 

#If the work is being done locally, the polling locations data lives here, as does the distance data and the outputs```

File: CLAUDE.md [hunk 13]
```diff
@@ -0,0 +1,95 @@
...
+```
+
+`run.py <script_name> [args]` → `docker compose run --rm app python -m python.scripts.<script_name> [args]`
+
+## Documentation Strategy
+
+This is an open source project. Keep the repo authoritative:
+
+- **Finalized decisions** → `docs/decisions/` (ADR format: "Why we chose X over Y")
```
Amasus: What is `doc/decisions`? Is this new?

Amasus:
There's something weird going on with `python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py` (see inline comments) is it possible this is deprecated?

CLAUDE.md
abd1tus: Yes it is new, the code generation develops a lot of spammy spec and plan markdown files.  They can fall out of date.  Instead, I looked into some alternatives and docs/decisions/ is a smaller summary for people working on the code on architectural decisions.

CLAUDE.md
abd1tus: Agreed will update

CLAUDE.md
abd1tus: Yeah technically, it has a __main__ function on it.  Unless you want to move it to a util folder somewhere.

CLAUDE.md
abd1tus: I'll update.

docs/to_install.md
abd1tus: Sounds good.

python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py
abd1tus: Yeah, I am still diving into these.

CLAUDE.md
Amasus: check to make certain that -e should be in model_run_cli as well.

update documentation in doc/to_run.md to show that this is optional.

also, is -l still needed?

CLAUDE.md
Amasus: actually, can we put pull_census_data under scripts?
also have it show up under run.py

CLAUDE.md
Amasus: move docs/decisions to docs/developer/decisions. Then we can put notes for debugging, etc in here as well.

python/database/alembic/versions/05a1ef156dca_rename_polling_locations_sets_to_.py
abd1tus: Changes to migration files are are as expected.

python/database/alembic/versions/5d5706c75d62_rename_polling_locations_only_to_.py
abd1tus: Changes to migration files are are as expected.

python/database/alembic/versions/5d5706c75d62_rename_polling_locations_only_to_.py
abd1tus: Changes to migration files are are as expected.

docs/to_install.md
abd1tus: Split out to docs/development/CONTRIBUTING.md

CLAUDE.md
abd1tus: -e and -l are both optional.  New command line args have been resolved in feature/readme_clarification.

CLAUDE.md
abd1tus: We now have python/scripts/pull_census_data_cli.py that shows up under run py.  The core utility still lives in python/utils/pull_census_data.py while the command line parsing now lives in  python/scripts/pull_census_data_cli.py.

CLAUDE.md
abd1tus: Updated.

CLAUDE.md
abd1tus: Moved to docs/development/decisions/  (instead of docs/developer/decisions)

docs/development/CONTRIBUTING.md
```diff
@@ -0,0 +1,118 @@
...
+1. Install conda if you do not have it already
+    1. This program uses SCIP as an optimizer, which is easily installed using Conda, but not using pip. (SCIP installation will be completed below by installing 'environment.yml')
+    1. If you do not have conda installed already, use the relevant instructions [here](https://conda.io/projects/conda/en/latest/user-guide/install/index.html)
+
+1. Create and activate conda environment. (Note, on a Windows machine, this requires using Anaconda Prompt.)
+    1. `$ conda env create -f environment.yml`
+    1. `$ conda activate equitable-polls`
+
+
+## Claude Code
```
Amasus: Is this necessary?
I feel like the others parts of this document are about how to run the program. This is about how to get a third party app to help with a workflow. We don't, for instance, put in instructions of how to use VSCODE or Gemini's Codex.

python/scripts/pull_census_data_cli.py
```diff
@@ -0,0 +1,26 @@
+"""Command line utility to pull census data for a given state and county."""
+
+import argparse
+
+from python.utils.pull_census_data import pull_census_data
+
```
Amasus, line 6: What's the logic in doing it this way instead of just moving `python.utils.pull_census_data` to `python.scripts.pull_census_data`

CLAUDE.md
Amasus: I asked this above. What is the logic behind keeping the function where it was and moving the command line parsing part of it out?

abd1tus: The core logic is still a utility used by other parts of the code.  The files in scripts are cli related, and the pull census code is also used by model_data.py.  It's a little janky to have it try to access the script cli code directly.

Amasus:
* When I run `python run.py e2e_tests -m e2e_csv` based on test plan above, I get ` 53 passed, 24 deselected, 19 warnings`
* When I run `python run.py e2e_tests -m e2e_db` based on test plan above, I get `24 skipped, 53 deselected, 19 warnings`
   * More importantly, it never asks me to connect to the database
* When I run `docker compose run --rm app pytest` based on test plan above, I get `7 failed, 122 passed, 24 skipped, 20 warnings`
* When I run `pylint python/` based on the test plan above, I get `'pylint' is not recognized as an internal or external command,
operable program or batch file.`

docs/development/CONTRIBUTING.md
abd1tus: This was marked as optional in the latest CONTRIBUTING.md.

python/scripts/pull_census_data_cli.py
abd1tus: We spoke about this on our last meeting.  Closing out.

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+1. Read this guide (especially [Getting Started](#getting-started) and [Code Style](#code-style))
+2. [Set up your development environment](#development-environment)
+3. Pick an issue or propose a change
+4. [Submit a pull request](#submitting-changes)
 
+Questions? [Ask us on Discord](https://discord.com/channels/1106301559811350540/1106301560507609241).
 
-# Code of Conduct
-Welcome to Voting Rights Code. Thank you for considering working with us. If you are just browsing, reach out. We'd love to talk to you. 
+## Code of Conduct
```
Amasus, line 16: Some of this is outdated. The rest of it is just "professionalism" do we still need it?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+python run.py model_run_cli -c NUM -l ./datasets/configs/<County>/config.yaml
+
+# Run tests
+docker compose run --rm app pytest
+
+# Run E2E tests
+python run.py e2e_tests
+```
+
+### Local Development with Conda (optional)
```
Amasus, line 167: Do we need to make it clear that you don't need this if you are using Docker?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+Inline comments should explain *why*, not restate *what* the code does. Keep them concise and follow PEP 8 (single space after `#`, sentence case).
+
+### R
+
+The R analysis scripts in `R/` follow the [tidyverse style guide](https://style.tidyverse.org/).
+
+- **Naming:** `snake_case` for all names. Files: `snake_case.R`
+- **Assignment:** Always `<-`, never `=`. Use `TRUE`/`FALSE`, never `T`/`F`
+- **Data manipulation:** This project uses `data.table` — use `dt[i, j, by]` idiom, not dplyr verbs. Do not mix.
+- **Vectorize operations:** Avoid explicit loops for element-wise work. Use `vapply()` over `sapply()`. Use `seq_along(x)` / `seq_len(n)`, never `1:length(x)`.
```
Amasus, line 253: This is not how things are done currently. Do you want me to change how I'm doing things?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+- **Linting:** Lint with `lintr::lint()`. Format with `styler::style_file()`.
```
Amasus, line 254: Also not done currently. Do you want me to start? 

Also, can the linting be automated in the CI/CD pipeline?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+- Run `alembic upgrade head` to initialize or update the database schema
+
+See [Database](docs/database.md) for full documentation on schema, imports, scratch dataset setup, and Alembic migrations.
+
+
+## Code Style
+
+### Python
+
+This project follows the [Google Python style guide](https://google.github.io/styleguide/pyguide.html), enforced via Pylint (see `.pylintrc`).
```
Amasus, line 214: Is the python pylint enforced? Is this a recent change? If so, that's great.

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+### Commit Messages
+
+This project uses [Conventional Commits](https://www.conventionalcommits.org/):
+
+```
+type(scope): description
+```
+
+- **Types:** `feat`, `fix`, `refactor`, `test`, `chore`, `docs`
+- **Scope:** Module name — `solver`, `database`, `scripts`, etc.
```
Amasus, line 265: This is not how branches are currently made. Again, is this a change you want to start making?

What if we have commits that affect multiple scopes?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+
+1. Ensure `settings.yaml` has a `test` environment configured (see `settings_example.yaml`)
+2. Ensure GCP Application Default Credentials are set up (`gcloud auth application-default login`)
+3. Run via `python run.py e2e_tests -m e2e_db` — the `run.py` wrapper mounts GCP credentials into the Docker container automatically
+
+DB tests clean up after themselves by deleting all `e2e_`-prefixed data from the test dataset at both the start and end of each session.
+
+### Test-Driven Development
+
+All code should be written using TDD:
```
Amasus, line 346: We don't do test driven development. I don't know that I want to, but am open to having my mind changed. Can we discuss?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+```
+
+All tests and lints must pass.
+
+
+## Linting
+
+### Python
+
+All code changes must pass Pylint before committing:
```
Amasus: This is not how I/ we do things. Again, is this a new standard you are trying to set up?

CONTRIBUTING.md
```diff
@@ -1,43 +1,444 @@
...
+When your change is ready:
+
+1. Ensure all tests pass (`python run.py test`)
+2. Ensure lint passes (`pylint python/`)
+3. Include tests for new features
+4. Write a clear PR description explaining what changed and why
+
+### Code Review
+
+All pull requests require review by at least **two maintainers** before merge.
```
Amasus, line 430: We only have two active maintainers.... Do you mean this?

abd1tus:
> * When I run `python run.py e2e_tests -m e2e_csv` based on test plan above, I get ` 53 passed, 24 deselected, 19 warnings`
> * When I run `python run.py e2e_tests -m e2e_db` based on test plan above, I get `24 skipped, 53 deselected, 19 warnings`
>   
>   * More importantly, it never asks me to connect to the database
> * When I run `docker compose run --rm app pytest` based on test plan above, I get `7 failed, 122 passed, 24 skipped, 20 warnings`

The "deselected" count is pytest's way of ignoring specific tests that don't apply - i.e. running csv tests and ignoring db tests.  The warnings can be cleaned up, but should be out of scope for e2e.  Ideally we will always get 0 warnings. 

The skipped is a a problem.  That means it was supposed to run tests but it decided not to because the DB could not be reached.  This is a design bug.  It will now throw an explicit error if it cannot connect instead of leaving the user with an ambiguous "24 skipped".  Good catch! 

I also updated the CONTRIBUTING.md docs to indicate that a properly configured "test" environment in settings.yaml is required. 

> * When I run `pylint python/` based on the test plan above, I get `'pylint' is not recognized as an internal or external command, operable program or batch file.`

pylint has not been added to environment.yaml.  You can access pylint via conda or ```python run.py lint```.  Additional instructions have been added to CONTRIBUTING.md about linting.

CONTRIBUTING.md
```diff
@@ -1,43 +1,494 @@
...
+
+## Development Environment
+
+### Docker (recommended)
+
+Docker is the primary development method. All commands go through `run.py`, which handles Python dependencies, SCIP optimizer setup, and environment configuration inside the container.
+
+```bash
+# Run the model locally
+python run.py model_run_cli -c NUM -l ./datasets/configs/<County>/config.yaml
```
Amasus: I think Claude pulled this from the erroneous line from claude.md.

this should be <config_set>/config.yaml with some indication that it accepts wild cards so <config_set>/*.yaml  works too

docs/result_analysis.md
```diff
@@ -87,12 +87,16 @@ And at the end of the R file, add the following line:
 upload_graph_files_to_cloud_storage()
 ```
 
-Authentication will happen via the [`googleCloudStorageR` R library](https://CRAN.R-project.org/package=googleCloudStorageR).  Before running the R analysis, you must login via the [gcloud command line tool](https://cloud.google.com/sdk/docs/install-sdk).  One installed, run the following:
+Authentication will happen via the [`googleCloudStorageR` R library](https://CRAN.R-project.org/package=googleCloudStorageR). Before running the R analysis you need GCP Application Default Credentials. gcloud is pre-installed in the dev container — run the following from the container's integrated terminal:
```
Amasus, line 92: Clarifying question:
This is making me think that if Joe R. GIS Director wants to use our open source code themselves, they would have to get access to OUR storage bucket. 

Is that the case? And if so is that desirable?
Not something we need to fix for this branch, but it should go on the list.

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _apply_log_transform(src_path: str, dest_path: str) -> None:
+    """Read a distance CSV and write a copy with log-transformed distance_m values.
+
+    Values that are zero or negative are left unchanged (log is undefined there).
```
Amasus: Negative distances should raise an error. 
I think Zero errors should to. I believe I've masked zero distances (especially in the driving case) to a small number to avoid this issue.

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+    df['distance_m'] = df['distance_m'].apply(lambda x: math.log(x) if x > 0 else x)
+    df.to_csv(dest_path, index=False)
+
+
+def _make_config(base: dict, sid: str, suffix: str, overrides: dict) -> dict:
+    """Return a config dict derived from *base* with e2e-specific fields applied.
+
+    Args:
+        base: The base configuration loaded from the template YAML.
+        sid: The session ID used as both ``config_set`` and ``location``.
```
Amasus, line 238: Why are we doing it this way (setting config name to sid) rather than using the available test configs?

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+
+    Returns:
+        A string of the form ``e2e_<6 hex chars>``.
+    """
+    return f'e2e_{uuid.uuid4().hex[:6]}'
+
+
+@pytest.fixture(scope='session')
+def e2e_test_data(e2e_session_id):
+    """Create isolated test data directories and files for the e2e session.
```
Amasus, line 270: Same question as above, why create isolated test data directories and files, instead of using the pre-existing ones?

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+        with open(config_path, 'w', encoding='utf-8') as fh:
+            yaml.dump(cfg, fh, default_flow_style=False, allow_unicode=True)
+        config_paths[suffix] = config_path
+
+    # --- Autogen template ----------------------------------------------------
+
+    autogen_template = dict(base_config)
+    autogen_template['config_set'] = sid
+    autogen_template['config_name'] = f'{sid}_autogen'
+    autogen_template['location'] = sid
```
Amasus, line 388: Will this work? 
Where is the sid census data created? 
Which census pull is it this based off of?
Am I misunderstanding what is meant by end to end?

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+
+    # Teardown — remove the directories unconditionally.
+    for dirpath in (polling_subdir, driving_subdir, config_subdir):
+        if os.path.isdir(dirpath):
+            shutil.rmtree(dirpath)
+
+
+@pytest.fixture(scope='session')
+def test_environment():
+    """Load the 'test' environment from settings.yaml.
```
Amasus, line 432: Is this something we need to set up? my settings.yaml doesn't have this currently. Perhaps this should throw a flag indicating that there isn't the appropriate entry in the file. Else, a bunch of tests get skipped, and I don't know why.

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+    # up front instead of mid-test with confusing downstream errors.
+    engine = sqlalchemy_main.setup(env)
+    engine.connect().close()
+
+    yield env
+
+
+@pytest.fixture(scope='session')
+def clean_test_data(test_environment):
+    """Delete all e2e-prefixed rows from the DB before (and after) the session.
```
Amasus: Why are these being deleted?
Why not just keep a separate test env with various test results around?
In fact, I wanted to talk to you about having various test data up and around in the db, anyway, because it would make certian bits of feature development easier.

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+    pl_set_id = pl_set.id
+
+    driving_set = query.find_driving_distance_set(census_year, '20250101', sid)
+    driving_set_id = driving_set.id if driving_set else None
+
+    permutations = [
+        # (log_distance, driving, csv_key)
+        (False, False, 'distances_import'),
+        (True, False, 'distances_log_import'),
+        (False, True, 'driving_distances_import_dd'),
```
Amasus, line 558: what is the "dd" suffix?

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+    e2e_test_data,
+    clean_test_data,
+    test_environment,
+    imported_potential_locations,
+    imported_driving_distances,
+):
+    """Import all four distance-data permutations for the e2e session into the DB.
+
+    Rather than calling ``db_import_distance_data_cli`` (which rebuilds distance
+    data from Census Tiger shapefiles — a cross-join producing ~500 MB for a
```
Amasus, line 515: Again, wondering why we are creating, rather than using the census data that is already there in the test/ folders.

python/tests/e2e/test_auto_generate_config.py
```diff
@@ -0,0 +1,211 @@
...
+from python.tests.e2e.conftest import run_cli
+
+MODULE = 'python.scripts.auto_generate_config'
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _generated_config_paths(e2e_test_data: dict) -> list[str]:
```
Amasus, line 24: Again, why is sid being used for location, and not "testing", as we've been doing so far?

python/tests/e2e/test_auto_generate_config.py
```diff
@@ -0,0 +1,211 @@
...
+    sid = e2e_test_data['sid']
+    config_dir = e2e_test_data['config_dir']
+    return [
+        os.path.join(config_dir, f'{sid}_year_2020.yaml'),
+        os.path.join(config_dir, f'{sid}_year_2022.yaml'),
+    ]
+
+
+# ---------------------------------------------------------------------------
+# CSV tests
```
Amasus, line 47: YAML tests?

python/tests/e2e/test_auto_generate_config.py
```diff
@@ -0,0 +1,211 @@
...
+
+        Args:
+            e2e_test_data: Session-scoped test data dict.
+        """
+        template_path = e2e_test_data['autogen_template']
+
+        run_cli(MODULE, '-b', template_path)
+
+        expected_years = {
+            '_year_2020.yaml': ['2020'],
```
Amasus, line 104: I believe that this is the right format for the database, but the local yaml files aren't lists in the same way. Will this test for local creation as well?

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def test_driving_dir_exists(self, e2e_test_data):
+        """The driving sub-directory must exist."""
+        assert os.path.isdir(e2e_test_data['driving_dir'])
+
+    def test_config_dir_exists(self, e2e_test_data):
+        """The config sub-directory must exist."""
+        assert os.path.isdir(e2e_test_data['config_dir'])
+
+    def test_potential_locations_csv_exists(self, e2e_test_data):
```
Amasus, line 56: Now I'm confused, where is it getting the csvs from?

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def test_driving_distances_csv_exists(self, e2e_test_data):
+        """The driving distances CSV (in polling dir) must exist."""
+        assert os.path.isfile(e2e_test_data['driving_distances'])
+
+    def test_driving_distances_log_csv_exists(self, e2e_test_data):
+        """The log-transformed driving distances CSV must exist."""
+        assert os.path.isfile(e2e_test_data['driving_distances_log'])
+
+    def test_driving_distances_import_csv_exists(self, e2e_test_data):
```
Amasus, line 76: why is this a different csv than the one in `e2e_test_data['driving_distances']`?

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+        """Exactly 8 config variants must be created."""
+        assert len(e2e_test_data['configs']) == 8
+
+
+@pytest.mark.e2e
+@pytest.mark.e2e_csv
+class TestE2eTestDataContent:
+    """Verify that the generated files have the expected content."""
+
+    def test_potential_locations_csv_is_non_empty(self, e2e_test_data):
```
Amasus: I don't like the creating of these and checking that these are non-empty on the fly. Specifically, this gives me no insight into what these datasets are or how they are created, or how I should interpret the results. 

The way the rest of the tests are set up, there is a location, called testing, and a sample of the Gwinnett, GA data (tens of rows) that live  in the correct census files. There is also a set of associated potential location data and a bunch of configs (maybe not enough, but if so, these just need to be expanded upon). Everything but "pull_census_data" can and should be tested off these existing sets. 

I think these sets should be uploaded to a test branch somewhere in the db, and kept around locally. That way they can be used for feature development easily.

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+    def test_log_distances_have_distance_m_column(self, e2e_test_data):
+        """The log-transformed distances CSV must have a 'distance_m' column."""
+        df = pd.read_csv(e2e_test_data['distances_log'])
+        assert 'distance_m' in df.columns
+
+    def test_log_transform_applied_to_distances(self, e2e_test_data):
+        """Log-transformed distances must differ from the linear source distances."""
+        src_df = pd.read_csv(e2e_test_data['distances'])
+        log_df = pd.read_csv(e2e_test_data['distances_log'])
+        # At least some values should differ after log transformation.
```
Amasus: log has no fixed points. ALL values should differ after the log transformation.

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+        positive_mask = src_df['distance_m'] > 0
```
Amasus: I believe that distance_m should always be greater than 0. If not, we need to mask it in the code, and check that this is the case in the tests, not mask for it here.

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+        if positive_mask.any():
+            assert not src_df.loc[positive_mask, 'distance_m'].equals(
+                log_df.loc[positive_mask, 'distance_m']
+            )
+
+    def test_config_set_matches_session_id(self, e2e_test_data):
```
Amasus, line 132: See previous concerns about session id.

python/tests/e2e/test_conftest_fixtures.py
```diff
@@ -0,0 +1,199 @@
...
+        """Every generated config YAML must have config_set equal to the session ID."""
+        sid = e2e_test_data['sid']
+        for suffix, config_path in e2e_test_data['configs'].items():
+            with open(config_path, 'r', encoding='utf-8') as fh:
+                cfg = yaml.safe_load(fh)
+            assert cfg['config_set'] == sid, (
+                f"config_set mismatch in '{suffix}': expected {sid!r}, got {cfg['config_set']!r}"
+            )
+
+    def test_config_location_matches_session_id(self, e2e_test_data):
```
Amasus, line 142: see previous concern about session id.

python/tests/e2e/test_db_import_distance_data.py
```diff
@@ -0,0 +1,235 @@
...
+            location=sid,
+            log_distance=False,
+            driving=False,
+        )
+        assert ds is not None, (
+            f"No DistanceDataSet found for location='{sid}', log_distance=False, driving=False"
+        )
+
+        db_df = query.get_distance_data(ds.id)
+        assert len(db_df) > 0, (
```
Amasus: again, we should know enough about the test data to be able to fix these sizes.

python/tests/e2e/test_db_import_distance_data.py
```diff
@@ -0,0 +1,235 @@
...
+            location=sid,
+            log_distance=False,
+            driving=False,
+        )
+        assert ds is not None, (
+            f"No DistanceDataSet found for location='{sid}', log_distance=False, driving=False"
+        )
+
+        db_df = query.get_distance_data(ds.id)
+        assert len(db_df) > 0, 'Expected at least one row of distance data'
```
Amasus: Again I think we should know enough about the distance data that we should know what the length is. 

Maybe the problem is that I don't understand what data is being pushed up to the database for this in the first place. Is it the distance data that is sitting around in datasets/polling/testing? 

the entire sid thing is confusing me.

python/tests/e2e/test_db_import_distance_data.py
```diff
@@ -0,0 +1,235 @@
...
+            f"No DistanceDataSet found for location='{sid}', log_distance=False, driving=False"
+        )
+
+        db_df = query.get_distance_data(ds.id)
+        assert len(db_df) > 0, (
+            'Expected at least one DistanceData row for the linear haversine permutation'
+        )
+
+    def test_demographic_columns_populated(self, e2e_test_data, imported_distance_data_all, test_environment):
+        """The population, white, black, and hispanic columns are populated in the DB.
```
Amasus: why are we looking for only these columns?

python/tests/e2e/test_db_import_distance_data.py
```diff
@@ -0,0 +1,235 @@
...
+            location=sid,
+            log_distance=True,
+            driving=False,
+        )
+        assert ds is not None, (
+            f"No DistanceDataSet found for location='{sid}', log_distance=True, driving=False"
+        )
+
+        db_df = query.get_distance_data(ds.id)
+        assert len(db_df) > 0, (
```
Amasus: same concerns as above.

python/tests/e2e/test_db_import_distance_data.py
```diff
@@ -0,0 +1,235 @@
...
+Tests cover all four distance-data permutations: linear haversine, log haversine,
+linear driving, and log driving.  All tests are database-backed and will be
+skipped automatically when no 'test' environment is configured in settings.yaml.
+"""
+
+# pylint: disable=redefined-outer-name
+
+import pytest
+
+
```
Amasus: I feel like there is a lot of repeated code in the 4 cases below. Surely, there's a better way of writing these tests that don't involve repetition, and thus the need to fix each case individually when something changes (like getting rid of or changing the SID framework)

python/tests/e2e/test_db_import_driving_distances.py
```diff
@@ -0,0 +1,89 @@
...
+
+        Args:
+            e2e_test_data: Session-scoped test data dict.
+            imported_driving_distances: Ensures the import has run.
+            test_environment: The loaded test environment.
+        """
+        from python.database.query import Query  # pylint: disable=import-outside-toplevel
+
+        sid = e2e_test_data['sid']
+        csv_path = e2e_test_data['driving_distances_import']
```
Amasus: why is this not
e2e_test_data['driving_distances']  ?

python/tests/e2e/test_db_import_driving_distances.py
```diff
@@ -0,0 +1,89 @@
...
+            census_year='2020',
+            map_source_date='20250101',
+            location=sid,
+        )
+        assert distance_set is not None, (
+            f"No DrivingDistancesSet found for location '{sid}'"
+        )
+
+        db_df = query.get_driving_distances(distance_set.id)
+        assert len(db_df) > 0, 'Expected at least one driving distance row in DB'
```
Amasus: Again, we should know more about the dataset and fix the length.

python/tests/e2e/test_db_import_driving_distances.py
```diff
@@ -0,0 +1,89 @@
...
+        assert (db_df['distance_m'] > 0).all(), (
+            f"All distance_m values must be positive; found non-positive values:\n"
```
Amasus, line 85: This is good, and as it should be. However, the log distances tests do not assume this.

This inconsistency should be fixed.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+    }
+
+
+# ---------------------------------------------------------------------------
+# Cleanup fixture
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope='module', autouse=True)
+def cleanup_results(e2e_session_id):
```
Amasus: I've noted this before, but I think we shouldn't have the cleanups. Happy to discuss.

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+
+CONFIG_VARIANTS = {
+    'config_basic': {},
+    'config_driving': {'driving': True},
+    'config_log': {'log_distance': True},
+    'config_driving_log': {'driving': True, 'log_distance': True},
+    'config_penalty': {'penalized_sites': ['College Campus - Potential', 'Fire Station - Potential']},
+    'config_low_beta': {'beta': -1},
+    'config_capacity': {'capacity': 3},
+    'config_constrained': {'maxpctnew': 0.5, 'minpctold': 0.75},
```
Amasus: Can we change this name? It completely confused me when I encountered it later. Maybe `config_new_locations`?

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+# ---------------------------------------------------------------------------
+# Class: TestModelRunCliBasic — smoke tests
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+@pytest.mark.e2e_csv
+class TestModelRunCliBasic:
+    """Smoke tests verifying the CLI runs successfully and writes output files."""
```
Amasus: Noticing that the different smoke tests use different config paths, almost arbitrarily. Is that standard practice? It seems odd to me.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+# ---------------------------------------------------------------------------
+# Class: TestModelRunCliVariants — config variations
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+@pytest.mark.e2e_csv
+class TestModelRunCliVariants:
+    """Tests that each config variant runs end-to-end without error."""
```
Amasus, line 240: Are the config variants supposed to be a complete set of variants we are likely to use? If so, tell me, and I'll verify this. If not, what is the logic behind testing _THIS_ set of variations?

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+@pytest.mark.e2e_csv
+class TestModelRunCliValueAssertions:
+    """Value-level assertions on result CSV content produced by model_run_cli."""
+
+    def test_results_columns(self, e2e_test_data):
+        """The results CSV must contain id_orig, id_dest, and distance_m columns.
```
Amasus: Why these three columns only?

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+        _ensure_run(e2e_test_data, 'config_basic')
+        sid = e2e_test_data['sid']
+        results_path = _result_files(sid, 'config_basic')['results']
+
+        df = pd.read_csv(results_path)
+        for col in ('id_orig', 'id_dest', 'distance_m'):
+            assert col in df.columns, f"Expected column '{col}' in results CSV"
+
+    def test_all_residences_assigned(self, e2e_test_data):
+        """Every id_orig in the results must be a valid residence from the distances file.
```
Amasus: Not sure why this is being tested.

1. A better test would be to ascertain that every residence id_orig in the distance file appears in the result file.
2. This feels like a test of the optimizer, and not of our code.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+        source_ids = set(distances_df['id_orig'].astype(str))
+        result_ids = set(results_df['id_orig'].astype(str))
+
+        assert result_ids.issubset(source_ids), (
+            f"Result id_origs not a subset of source distances: "
+            f"{result_ids - source_ids}"
+        )
+
+    def test_no_duplicate_assignments(self, e2e_test_data):
```
Amasus: Not sure why this is here. It feels like a test of SCIP and not our code.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+        df = pd.read_csv(results_path)
+        duplicates = df[df.duplicated(subset=['id_orig'], keep=False)]
+        assert duplicates.empty, (
+            f"Duplicate id_orig entries found in results: "
+            f"{duplicates['id_orig'].tolist()}"
+        )
+
+    def test_precincts_open_count(self, e2e_test_data):
+        """The number of unique open precincts must not exceed precincts_open (3).
```
Amasus: 1. where is it getting (3) from? Here's where my confusion of where / how it is creating the input data is coming up to bite me.
2. Again, this feels like a test of SCIP and not our code.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+        results_path = _result_files(sid, 'config_basic')['results']
+
+        df = pd.read_csv(results_path)
+        unique_precincts = df['id_dest'].nunique()
+        assert unique_precincts <= 3, (
+            f"Expected at most 3 open precincts, got {unique_precincts}"
+        )
+
+    def test_ede_demographics_present_and_positive(self, e2e_test_data):
+        """The EDE file must have at least one row and positive y_EDE values.
```
Amasus: The EDE file should have at least 6 rows (for now).

This should be read off from the column names of the results file.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+
+        df = pd.read_csv(edes_path)
+        assert len(df) > 0, 'EDE file must have at least one row'
+        assert 'y_EDE' in df.columns, 'EDE file must have a y_EDE column'
+        assert (df['y_EDE'] > 0).all(), (
+            f"All y_EDE values must be positive; found non-positive: "
+            f"{df[df['y_EDE'] <= 0]['y_EDE'].tolist()}"
+        )
+
+    def test_residence_distances_count(self, e2e_test_data):
```
Amasus, line 451: These numbers should be fixed (and known) from the input test data.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+            e2e_test_data: Session-scoped test data dict.
+        """
+        _ensure_run(e2e_test_data, 'config_basic')
+        sid = e2e_test_data['sid']
+        res_path = _result_files(sid, 'config_basic')['residence_distances']
+
+        df = pd.read_csv(res_path)
+        assert len(df) > 0, 'Residence distances file must have at least one row'
+
+    def test_precinct_distances_count(self, e2e_test_data):
```
Amasus, line 493: This number should be known. Also, it feels like this can/ should be combined with the tests_precincts_open_count test.

The number of open precincts should be the same in both cases.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+        """
+        _ensure_run(e2e_test_data, 'config_penalty')
+        sid = e2e_test_data['sid']
+        files = _result_files(sid, 'config_penalty')
+
+        results_df = pd.read_csv(files['results'])
+        edes_df = pd.read_csv(files['edes'])
+
+        assert len(results_df) > 0, 'Penalty config results must have at least one row'
+        assert len(edes_df) > 0, 'Penalty config EDE output must have at least one row'
```
Amasus, line 583: exactly 6 rows.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+            e2e_test_data: Session-scoped test data dict.
+        """
+        _ensure_run(e2e_test_data, 'config_penalty')
+        sid = e2e_test_data['sid']
+        files = _result_files(sid, 'config_penalty')
+
+        results_df = pd.read_csv(files['results'])
+        edes_df = pd.read_csv(files['edes'])
+
+        assert len(results_df) > 0, 'Penalty config results must have at least one row'
```
Amasus, line 582: We should know what this number is.

python/tests/e2e/test_model_run_cli.py
```diff
@@ -0,0 +1,505 @@
...
+        assert len(edes_df) > 0, 'Penalty config EDE output must have at least one row'
+
+        for col in ('id_orig', 'id_dest', 'distance_m'):
```
Amasus: Why just these three columns?

python/tests/e2e/test_model_run_db_cli.py
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+
+# ---------------------------------------------------------------------------
+# Cleanup fixture
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(scope='module', autouse=True)
+def cleanup_csv_results(e2e_session_id):
```
Amasus: I don't like the idea of cleaning up the csvs.
But I also don't like the idea of creating a new folder for each time we run the tests.
Let's talk to figure out a better way? (or convince me I'm wrong?)

python/tests/e2e/test_model_run_db_cli.py
```diff
@@ -0,0 +1,206 @@
...
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.e2e
+@pytest.mark.e2e_db
+class TestModelRunDbCliCsvOutput:
+    """Tests verifying that model_run_db_cli writes correct CSV output files."""
+
+    def test_single_config_csv_output(self, e2e_test_data, imported_distance_data_all, imported_configs):
+        """Running a single DB config with -o csv produces the expected result file.
```
Amasus, line 103: Confused. This is testing the db_cli... why do we have -o csv?

I thought that the -o flag was gone (or at least deprecated) ... this was why there was the two different cli commands in the first place.

Amasus (commented):
Okay, I finally went through the e2e tests, and there's a lot I don't understand or object to.

* I like the creation of the test environment. 
   * This currently fails silently if the environment isn't set up. It should raise an error. I had no idea why these tests were being skipped until I read this code.
   * The test environment should also be documented where the settings.yaml is introduced.
   * I think files should be written to the test environment, and not cleaned up afterwards
* I don't like (or understand) how the SID -session ids- are getting data for the model to run on 
   * The location is currently being set to the SID. How is it getting census data in this case? Similarly, what potential locations are it using?
   * Previously (for other tests) I'd set up a testing census and potential locations data suite that was sampled from Gwinett, so that we could test everything. 
   * I'd love for the e2e tests to use this
   * I'd also love for this data to live in the tests environment, as I'd love to use is for feature development. I do so currently, just locally.
* I like the idea of CONFIG_VARIANTS
   * Is that list supposed to be a comprehensive list? If so, let me know, and I'll look more closely.
   * I really like how it is being used to make certain that certain outputs are different.
* There are a few dumb tests in here.
   * I don't like how it's testing log distances. 
   * I don't like how it is checking that a valid csv is being created by checking that it has at least one row.
   * Similarly, there are times when it only checks for the existence of a handful of columns in the data frame, and I'm not sure why it's chosen those. 
* It is unclear to me if this is, indeed the full set of coverage that one needs for this. Certainly some things feel like silly tests (when it tests for things that SCIP should be handling, for instance). But I'm not sure that everything is tested... need to put some thought into that. Maybe not for this feature.

python/tests/conftest.py
```diff
@@ -14,31 +12,69 @@
 from .constants import (
   TESTING_CONFIG_BASE, TESTING_CONFIG_KEEP, TESTING_CONFIG_EXCLUDE, TESTING_CONFIG_PENALTY,
   TESTING_CONFIG_PENALTY_UNUSED, TESTING_CONFIG_DRIVING, TESTING_POTENTIAL_LOCATIONS_PATH, MAP_SOURCE_DATE,
-  TESTS_DIR,
 )
 
 
 # Fixtures for import tests
+#
+# These CSVs are generated into a session-scoped tmp dir rather than committed
```
Amasus: no. I REALLY do want the test .csvs available locally. 

I used them for feature development, and don't want to have to recreate them every time. Please fight gitignore instead.

I'm happy to do the github fiddling on this one, but i feel strongly that the changes to this file should be reverted.

python/tests/migration_schema_test.py
```diff
@@ -0,0 +1,479 @@
+"""Tests for migration schema validator."""
+import os
```
Amasus, line 2: FYI, not reviewing this carefully, mostly because I don't understand the schema validator well.

Happy to discuss in detail with you.

python/tests/migration_schema_validator.py
```diff
@@ -0,0 +1,754 @@
+"""Symbolic schema state tracker for validating Alembic migration chains.
+
+Walks the migration chain in revision order, tracks table/column/view state,
```
Amasus, line 3: Similarly, not reviewing because i don't fully undertand the features being tested here.

python/CLAUDE.md
```diff
@@ -0,0 +1,56 @@
...
+## Key Conventions
+
+- Config paths are case-sensitive: use `Gwinnett_GA`, not `Gwinnett_Ga`
+- Two data sources: `csv` (local files) or `db` (BigQuery) — set in `PollingModelConfig`
+- Environment names are defined in `settings.yaml`
+- Git LFS required for large dataset files (distances, shapefiles)
+
+## Contributing
+
+- PRs require 2 maintainer reviews; lint before merging
```
Amasus: this is for Claude... do we really want 2 maintainers... i.e. both of us?

install_r_packages.R
```diff
@@ -0,0 +1,84 @@
...
+#
+# Local R install (outside the container): drop `sudo` and omit
+# `library='/usr/local/lib/R/site-library'`. Both exist because the dev
+# container installs packages into a root-owned system library; a local R
+# install typically uses a user-owned library that renv detects
+# automatically.
+
+options(repos = c(CRAN = "https://cloud.r-project.org"))
+
+packages <- c(
```
Amasus, line 25: I worry that I'm seeing a packages list in multiple places: here, R/smoke_test, in .devcontainer.

If we update one and not the others what will happen?

renv.lock
Amasus, line 1: not reviewing

Amasus (commented):
Everything reviewed but the contributing now.

python/tests/conftest.py
lackadaisicalchad: Code updated, test csvs are not available again.  The original change was due to the fact that the expected csvs were not committed properly to the git repo due to gitignore rules.  gitignore has been updated.

python/CLAUDE.md
lackadaisicalchad: This isn't needed.  Removed.

CONTRIBUTING.md
abd1tus: Yanked as per last meeting.

docs/result_analysis.md
abd1tus: Doable, yeah.  When we integrated the google cloud we made decisions a while ago that anyone who wants to look at the data would need to request access.  I will clarify this in the results_analysis.md doc.  If there are some datasets that we'd like to truly make public then we can come up with a process for doing when we meet up.

python/tests/e2e/conftest.py
abd1tus: Done

python/tests/e2e/conftest.py
abd1tus: Done, an error will be raised.

python/tests/e2e/conftest.py
abd1tus: It allows for creating several permutations of configs based on available ones.  This code is called bellow where base values are loaded in from _SRC_BASE_CONFIG (which is testing_config_no_bg.yaml').  Then sid is just a short serial number to allow the config to be tracked as a session id (as stated).  This saves us from writing up about a half dozen or so hard coded configs with various kinds of edge cases and errors.  - see CONFIG_VARIANTS python/tests/e2e/conftest.py

python/tests/e2e/test_auto_generate_config.py
abd1tus: Yes, it is correct because it is validating against parsed yaml which has the year show up as an array of strings.

python/tests/e2e/test_model_run_cli.py
abd1tus: Going with:

[Notes]

31 = `python/tests/e2e/test_model_run_cli.py:318` (aka  def test_no_duplicate_assignments(self, e2e_test_data)  )

32 = `python/tests/e2e/test_model_run_cli.py:336`

Option B — drop #32 entirely if #31's matching invariant lands.

  If #31's proposed invariant ("matching column is binary {0,1} and sums to precincts_open") is in place, then
  df['matching'].sum() == cfg['precincts_open'] already implies the chosen-site count is what the config said. #32
  becomes redundant.

  Caveat I should be honest about: I don't actually know yet what matching represents on a per-residence row. If
  matching=1 means "this residence is assigned to this site", then df['matching'].sum() == num_residences, not
  precincts_open, and #31's proposed invariant would itself need a tweak (e.g., df.loc[df['matching']==1,
  'id_dest'].nunique() == cfg['precincts_open']). Worth verifying against the writer code before committing to either
  #31 or #32's replacement.

R/CLAUDE.md
abd1tus: @Amasus will think on this one.

CONTRIBUTING.md
abd1tus: Local hooks will be a better option.  Claude's response:

Yes — and for a project this size it might be the better starting point. Three pieces:

  Standard tool: pre-commit

  The framework is pre-commit.com. One config file .pre-commit-config.yaml checked into the repo, devs run pre-commit install once after cloning, hooks fire on git
  commit. Pylint has an official hook via mirrors-pylint. Sketch:

  # .pre-commit-config.yaml
  repos:
    - repo: https://github.com/pylint-dev/pylint
      rev: v3.x.x
      hooks:
        - id: pylint
          files: ^python/
          args: ['--rcfile=.pylintrc']

  Plus optional hooks: trailing-whitespace, end-of-file-fixer, check-yaml, etc. — all are zero-config.

CONTRIBUTING.md
abd1tus: Fixed

python/tests/e2e/test_model_run_cli.py
abd1tus: Done.

CONTRIBUTING.md
abd1tus: Removed.

CONTRIBUTING.md
Amasus: @abd1tus , this has been updated in the new branch I'm about to give you a PR on.

R/CLAUDE.md
Amasus: @abd1tus , this has been addressed in the new branch I'm about to make a PR for.

CONTRIBUTING.md
Amasus: @abd1tus , this has been updated in the new branch I'm about to give you a PR on. 

Except for linting. As per the discussion yesterday, you were going to use hooks to get automated linting set up.

docs/result_analysis.md
Amasus: IIRC, the question on this front was whether or not R/result_analysis.R _could_ be run without db access. Ideally, would want that to be the case. Probably scope creep for this PR. Can we deal with this separately?

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,636 @@
...
+_SRC_POTENTIAL_LOCATIONS = os.path.join(_TESTING_POLLING_DIR, 'testing_potential_locations.csv')
+_SRC_DISTANCES = os.path.join(_TESTING_POLLING_DIR, 'testing_distances_2020.csv')
+_SRC_DRIVING_DISTANCES = os.path.join(_TESTING_POLLING_DIR, 'testing_driving_2020.csv')
+_SRC_BASE_CONFIG = os.path.join(_TESTING_CONFIG_DIR, 'testing_config_no_bg.yaml')
+
+# ---------------------------------------------------------------------------
+# Config variants
+# ---------------------------------------------------------------------------
+
+CONFIG_VARIANTS = {
```
Amasus, line 78: Confused. I thought these variants corresponded to the config files that already exists in `datasets\configs\testing`. There are 7 files in there, and 8 configs here.

1. How is this choice of variants being made?
2. If this set of variants do correspond to the files in `datasets\configs\testing`, can we keep the naming convention such that it is easy to see what the corresponding file is?
3. If these variants do NOT correspond to existing config files, where / how can we see what these variants are?

Long story short, this section needs better documentation so that a human can follow and trace.

python/tests/e2e/test_model_run_db_cli.py
```diff
@@ -0,0 +1,213 @@
...
+
+        query = Query(test_environment)
+
+        # Confirm the ModelConfig record exists.
+        db_config = query.find_model_configs_by_config_set_and_config_name(sid, config_name)
+        assert db_config is not None, (
+            f"No ModelConfig found for config_set='{sid}', config_name='{config_name}'"
+        )
+
+        # Confirm at least one ModelRun was created for this config.
```
Amasus: Question:
Should this be at least one or exactly 1?

Amasus (dismissed):
Just a couple more clarifying questions. 

this is looking better

python/tests/e2e/test_model_run_cli.py
abd1tus: Updated.  It now dynamically checks all columns.

CONTRIBUTING.md
abd1tus: Merged in.

python/tests/e2e/test_model_run_db_cli.py
abd1tus: We discussed this in the last meeting.  There is the option to not clean up.

python/tests/e2e/test_db_import_distance_data.py
abd1tus: updated, it now checks for a fixed value

python/tests/e2e/test_db_import_distance_data.py
abd1tus: code updated - it now dynamically checks all columns

python/tests/e2e/conftest.py
```diff
@@ -0,0 +1,600 @@
...
+            f'Configure a \'test\' environment before running -m e2e_db. '
+            f'See CONTRIBUTING.md: Setting Up DB Tests.',
+            returncode=1,
+        )
+
+    # Mixed run: emit a prominent banner. Tests still skip individually.
+    bar = '=' * 80
+    msg_lines = [
+        bar,
+        f'DB e2e tests will be skipped: {len(db_items)} tests marked e2e_db, '
```
Amasus, line 184: Am I understanding correctly that this function ONLY creates test data sets for importing to db, and no more? 

If that is so, it needs to be clarified in the documentation.

Amasus (approved):

PR #221, opened 2026-05-22
Feature/auth credentials

abd1tus:
Better token management for pull_census_data.py.


Amasus (commented):
This looks good. 

Quick question before I approve:

I don't understand how changing the secrects from a constant in a *.py file to an entry in a .json file changes the issue of the authentications/ folder being blown away periodically.

I believe that the changes you have made fixes this. I just don't understand how.

Amasus:
Follow-up: `rdh_username` and `rdh_password` (used for Redistricting Data Hub access, currently plaintext-only fields in `authentication_files/credentials.json`) should be added to the `SECRETS` registry in `secret_store.py`, alongside `census`. That would wire them into `run.py secret set/get/clear/restore` and give them a keyring backup, consistent with how the census API key is handled.

Issue #283, opened 2026-06-26
[Precincts] assign_block_to_dominant_precinct() can show real, populated precincts as zero population


Amasus:
Related to #266. (Note: `assign_block_to_dominant_precinct()` itself was introduced in #267.

## Bug

While generalizing the Monongalia zero-population-precinct-drop logic for #282, found that `assign_block_to_dominant_precinct()` (in `R/result_analysis/utility_functions/shape_extraction_functions.r`) reports **`Monongalia_73` as having zero total population** -- but this does not appear to be a real zero, unlike the confirmed-empty `Monongalia_A` (sliver, ~7,939 sq m) and `Monongalia_B` (real-but-uninhabited floodplain strip, ~331,434 sq m, see #281).

`Monongalia_73`:
- Has a real area of ~621,510 sq m -- larger than `Monongalia_B`'s confirmed-real area.
- Is assigned to Skyview Elementary School, along with five other small precincts (64, 68, 71, 72, 73, 75) that all serve the same polling location.
- Shows 0 total population in `block_precinct_assignment`.

## Suspected root cause

`assign_block_to_dominant_precinct()` assigns each census block to exactly one "dominant" (largest-overlap) precinct -- winner-take-all. When several small precincts compete for overlap with the same blocks (as the six Skyview-cluster precincts do), it's plausible that no single block ever has precinct 73 specifically as its single largest-overlap precinct, even though 73 has real area and likely real registered voters. That would make the reported "zero population" an artifact of the dominant-block heuristic, not a true reflection of the precinct's population.

This has not yet been confirmed against the source population data directly (e.g. via a population-weighted-by-overlap-area calculation) -- only the area/polling-assignment facts above are confirmed.

## Why this blocks #282

#282 proposes dropping precincts with zero total population (per `block_precinct_assignment`) as the generalized fallback when a county doesn't provide its own precinct data. If this bug is real, that rule would incorrectly drop `Monongalia_73` (and potentially other small precincts sharing a polling location with neighbors), not just genuinely empty ones like `A`/`B`.

## Possible directions (not yet decided)

- Apportion each block's population across *all* overlapping precincts proportional to overlap area, rather than assigning the whole block to one dominant precinct.
- Don't auto-drop zero-population precincts at all -- flag them for human review (consistent with #282's existing principle of erroring/flagging rather than guessing for anything that needs human judgment).

Amasus:
**Root cause confirmed against real data.**

[...]

Issue #282, opened 2026-06-26
[Precincts] Generalize county-provided precinct/polling-location reconciliation into reusable functions


Amasus:
Follow-up to #268. The Monongalia work in #268/#281 reconciled a county-provided precinct-to-polling-location file against the state precinct shapefile entirely by hand, in scratch code in `R/result_analysis/extract_precincts.r`. This ticket generalizes that into reusable functions in `R/result_analysis/utility_functions/shape_extraction_functions.r` so any WV county with its own precinct-to-location data gets the same reconciliation without rebuilding it from scratch.

## Inputs

- The county-provided precinct/polling-location table (may be in wide format: one row per polling place, one or more columns listing the precincts it serves, or may have one row per precinct).
- A vector of column names identifying which columns hold the per-location precinct list (generalizes `Precincts_by_Location.csv`'s repeated `Prec` columns).
- The polling-place name and address columns.
- `county_precincts` as already extracted from the state shapefile (`Precinct_I`, `USER_POLL_`, per `extract_county_precincts()`).

## Behavior (generalizes #268's manual steps)

1. Reshape the county data from one-row-per-location to one-row-per-(location, precinct) -- generalizes the `melt()` step. (only needed if the county-provided data is wide)
2. Build a `Precinct_I`-style key from each row's precinct number, using the same county-name prefix convention as the state shapefile (`extract_county_precincts()` already establishes `COUNTY_NAME`).
3. Case-insensitively match each county location name against the state's `USER_POLL_` naming.
4. Merge the reconciled county data against `county_precincts` by (name, `Precinct_I`).

## What can't be automated -- must error, not guess

Steps 4 and 5 from the original outline (identifying name mismatches and resolving precinct/location disagreements) require human judgment -- the function should never silently guess, and there is no manual-override-table input. Instead:

- **Unresolved name mismatch**: a county location name with no case-insensitive match against `USER_POLL_` -> `stop()` with an error listing the unmatched name(s) and instructing the user to edit the names directly in the county-provided source file until they match the state's naming. Corrections live in the source data itself, not in a side-table that can drift from it.
- **Unresolved precinct/location mismatch**: a `Precinct_I` that's county-only or state-only after the merge (the #281-style disagreement) -> `stop()` with an error listing the mismatched precinct IDs/locations and instructing the user to investigate with the county/state and encode the resolution (as in #281) before re-running.

Only once both checks pass cleanly does the function return the merged/resolved precinct data.

[...]

Issue #281, opened 2026-06-26
[Precincts] Resolve Monongalia precinct 44/74 and 2/A/B per client feedback (interim, pending SoS follow-up)


Amasus:
Part of milestone "Monongalia County precinct extraction & distance flagging". Follow-up to #268.

## Client feedback received

> Regarding precincts 44 and 74, both precincts currently vote at the Granville Social Hall which is the correct information we gave you. The precincts were previously several elections ago in the two locations you mentioned. Use the data we provided. I will need to check with the Secretary of State to find out why their mapping is not current with our voter registration system.
>
> Regarding Morgantown High School. It is split into two precincts only by alphabet due to the large number of registered voters at the location. For all intent and purposes it is noted as precinct 2.

## Interim resolution applied in code (scratch work, `R/result_analysis/extract_precincts.r`)

1. **Precinct 44**: state shapefile's `USER_POLL_` overridden from `GRANVILLE FIRE DEPT BINGO HALL STATION 2` to `GRANVILLE SOCIAL HALL` (matches 74). The state's Bingo Hall assignment is confirmed stale/outdated.
2. **Precinct B**: geometry unioned into `Monongalia_2` (both go to Morgantown High School per the client's "noted as precinct 2" framing).
3. **Precinct A**: dropped. It's a sliver (~7,939 sq m vs. precinct 2's ~4.6M sq m and B's ~331k sq m — see `precinct_analysis_outputs/Monongalia_County_WV/precinct_2_a_b_split.png`), doesn't touch precinct 2, and the client's feedback gives no indication it should be folded in anywhere.

Net effect: 44 state precincts -> 42 after this resolution (A dropped, B merged into 2).

## Why this issue stays open

This is an **interim** resolution based on the client's verbal explanation, not a confirmed reconciliation with authoritative data. The client is independently following up with the Secretary of State to find out why the state's voter-registration-system mapping disagrees with the county's own records for precinct 44/74. Until that comes back:

- Don't treat precinct A's drop as final if the client's SoS follow-up surfaces a reason it should be retained or reassigned.
- Don't generalize this 44/74-style "state data is stale" resolution to other counties without similar client confirmation.

**This milestone (Monongalia County precinct extraction & distance flagging) cannot be closed until this is resolved with the client/Secretary of State.**

Amasus:
Update: checked population via `assign_block_to_dominant_precinct()` (block-level census data) for precincts 2, A, and B:

[...]

Issue #268, opened 2026-06-20
[Precincts] Resolve polling-location name mismatches between precinct shapefile and potential_locations.csv


Amasus:
Part of milestone "Monongalia County precinct extraction & distance flagging". Depends on #266.

The precinct shapefile's `USER_POLL_` field (assigned polling location name, upper case) does not always exactly match the `Location` column in `datasets/polling/Monongalia_County_WV/Monongalia_County_WV_potential_locations.csv`. Of 24 distinct polling locations, ~3-4 differ, e.g.:

- `BOPARC SENIOR/COMMUNITY CENTER` vs `BOPARC Senior Recreation Center`
- `ST MARY'S CATHOLIC CHURCH` vs `St. Mary's Roman Catholic Church`

**Update — these next two are not simple spelling differences, they're substantive disagreements between the client's file and the state's file about which precincts map to which locations.** Reached out to the client about both; we'll need to proceed with code knowing the mapping may change once they respond:

- **Granville Social Hall**: the client's file has it serving precincts 44 and 74. The state file has it serving only 74, and assigns a separate location — Granville Fire Department Bingo Hall (Station 2) — to precinct 44. The client's file doesn't list the Fire Department Bingo location at all.
- **Morgan(town) High School**: the client's file has it serving precincts 2A and 2B. The state file has it serving precinct 2 (no letter suffix), and separately lists precincts A and B (no number) with no polling location assigned to either.

Human-review all of these and build an explicit manual lookup table (precinct name -> potential_locations name). Exact case-insensitive string match covers the rest not listed above. No fuzzy/automated matching. For the two substantive disagreements, the lookup table will need to encode whatever resolution comes back from the client — don't guess at one now.

Amasus:
Client feedback on discrepancies between polling locations:

> Regarding precincts 44 and 74, both precincts currently vote at the Granville Social Hall which is the correct information we gave you. The precincts were previously several elections ago in the two locations you mentioned. Use the data we provided. I will need to check with the Secretary of State to find out why their mapping is not current with our voter registration system.
> 
> Regarding Morgantown High School. It is split into two precincts only by alphabet due to the large number of registered voters at the location.  For all intent and purposes it is noted as precinct 2.

antisocialscientist:
This has been resolved through the forced manual edits of the county files, instead of a lookup table, correct?

Amasus:
[...]

PR #284, opened 2026-06-27
feat(precincts): generalize county-provided precinct reconciliation (#282)

Amasus:
## Summary

Generalizes #268's manual Monongalia name-matching into reusable functions, per #282:

- `R/result_analysis/utility_functions/shape_extraction_functions.r`: new `reshape_county_precincts_long()`, `add_precinct_id()`, `match_location_names()`, `reconcile_precinct_ids()`, `reconcile_county_provided_precincts()`. These reconcile a county-provided precinct/polling-location file against the state-extracted `county_precincts`, erroring with actionable instructions on any unresolved name or precinct-ID mismatch rather than guessing or silently dropping.
- `R/result_analysis/extract_precincts.r`: wires the above behind a new `COUNTY_PROVIDES_PRECINCT_DATA` config flag. Both branches (`TRUE`/`FALSE`) now return a `county_precincts_resolved` object with the same shape (`sf`, columns `Precinct_I`/`County_Nam`/`USER_POLL_`/geometry) for Step 6 to consume identically regardless of which path ran.
- `R/result_analysis/Extraction_configs/Monongalia_County_WV.r` (+ new `Monongalia_County_WV_no_precinct_data.r` for testing the `FALSE` branch): adds the new config variables.
- `R/result_analysis/scratch_correct_state_precinct_names.r` and `scratch_correct_state_precinct_ids.r`: one-off scripts used to correct Monongalia's source state shapefile (3 naming mismatches, then the 44/A/B/2-split precinct-ID disagreements per #281's client-confirmed resolution), each backing up the prior version before editing in place.
- `R/result_analysis/scratch_precinct_2_a_b_plot.r`: standalone visual-inspection script for the precinct 2/A/B split, used while investigating #281.

`EXPECTED_PRECINCT_COUNT`'s `stopifnot()` check is currently commented out in `extract_precincts.r` -- see the inline comment; #283 (dominant-block population heuristic can show real, populated precincts as zero) is still open and affects the precinct count.

## Review focus requested

This code was written by Claude (AI pair-programming assistant), so please review with extra scrutiny on:

- **Correctness and consistency with the associated tickets** (#268, #281, #282, #283 -- linked inline in code comments).
- **Coding style** -- it may not match the conventions used elsewhere in this codebase.
- **Code efficiency** -- is there a more idiomatic/efficient way to write any of this logic (data.table usage, sf handling, etc.)?
- **Comment clarity** -- comments reference repo issue numbers in places; please check whether they're still understandable to a reader unfamiliar with those tickets. (References to #283 specifically are fine to leave as-is for now, since that bug is still open and tracked separately.)
- **Error message clarity** -- error messages may need to be reformatted for a user to better understand what to change and how in the input data. suggestions welcome.
## Test plan

- [x] Manually verified both `COUNTY_PROVIDES_PRECINCT_DATA` branches produce a `county_precincts_resolved` object with matching shape and class.
- [x] Verified `reconcile_county_provided_precincts()` correctly errors on name mismatches (caught 3 real Monongalia naming differences) and precinct-ID mismatches (caught the 44/A/B/2-split disagreements), with messages confirmed against real data at each step.
- [x] Verified the full `extract_precincts.r` script runs end-to-end without error after the state shapefile corrections.
- [ ] No automated tests added yet (R project has no formal test suite per `R/CLAUDE.md`; manual verification only).


R/result_analysis/utility_functions/shape_extraction_functions.r
antisocialscientist, line 1: The comments are very verbose. Is that intentional? I understand intentionally verbose, but the explanations about things like unbuffered boundaries would be confusing to someone who doesn't know that path of development

R/result_analysis/extract_precincts.r
antisocialscientist, line 1: confirming that this is a flag for right now, as per the ticket, and not something that people in the future just need to be aware of. If the latter, referring to the specific ticket would be confusing
  # No county-provided data to reconcile against: fall back to dropping
  # precincts with zero population, per block-level population data in
  # block_precinct_assignment (#283 caveat: the dominant-block population
  # heuristic can show real, populated precincts as zero -- e.g.
  # Monongalia_73 -- so this fallback isn't fully trustworthy yet)

antisocialscientist (commented):
Other than a couple of comments on the comments, this looks good to go

Amasus:
Thanks for the review — addressed both:

1. **Verbose comments**: trimmed several, including the "(unbuffered boundary)" reference you flagged — that was leftover context from development that wouldn't mean anything to a reader unfamiliar with the implementation history. Also shortened a few others in `shape_extraction_functions.r` that explained reasoning that's now self-evident from the code/error messages themselves.
2. **#283 caveat scope**: confirmed it's a temporary flag, not a permanent caveat — it goes away once #283 (dominant-block population heuristic) is fixed. Will update the comment to say so explicitly.

Beyond those two comments, also found and fixed while testing this further:
- Generalized the hardcoded `"Prec"` county-column detection into a config value (`COUNTY_PRECINCT_COLUMN_NAMES`), with a count-validation check so a config typo fails loudly.
- Renamed `county_precincts` → `state_precincts` throughout (it held state-shapefile data, not county-provided data — confusing next to `provided_polls`), and renamed `reconcile_precinct_ids()` → `check_precinct_id_agreement()`.
- Confirmed the pipeline handles a long-format county-provided file (one row per precinct) with a config change only, no code changes.
- Added test configs to deliberately exercise both `stop()` error paths (name mismatch, ID mismatch) against the uncorrected state-shapefile snapshots, and confirmed both fire correctly.
- Improved the ID-mismatch error message to say which side (state vs. county) each disagreeing pair is missing from.
- Removed a duplicate `extract_county_precincts()` definition.

For the next review, can you:
- See if there are any other stylistic changes that can / should be made to make the code more readable (like the name changes above). Including, but not limited to:
   - function refactoring / reordering in shape_exctaction
   - more or different inline commenting.
   - anything that is currently done that isn't "R-like"

Amasus:
**Follow-up: refactor + a real correctness bug fix**

- Pulled Step 5's county-provided-data branch (column detection, slot-count validation, renaming, reconciliation call) into a single new helper, `reconciled_state_precinct_data()` — `extract_precincts.r`'s `if` block is now one function call instead of ~25 lines.
- Found and fixed a behavior mismatch between the two branches: when the county provides precinct data, zero-population precincts were being kept; when it doesn't, they were dropped. Restructured so the population-zero filter runs once, identically, after either branch produces its result — so both paths now apply the same rule.
- Verified the fix: wrote a one-off comparison script (`scratch_compare_precinct_resolution_paths.r`) that runs both the `COUNTY_PROVIDES_PRECINCT_DATA = TRUE` and `= FALSE` paths against the same Monongalia data and confirms they now produce identical results (41 rows each, matching `Precinct_I`/`County_Nam`/`USER_POLL_`).

R/result_analysis/utility_functions/shape_extraction_functions.r
antisocialscientist, line 1: Expand comment on lines 117-120 explaining why you are only checking one way.

antisocialscientist, line 1: 142-145: Again, clarify that you are intentionally only looking in one direction

antisocialscientist:
You are right, these are much better error messages. I do think a bit of explanation of why you are only checking in one direction (because you take the county as ground truth) would clarify.

R/result_analysis/utility_functions/shape_extraction_functions.r
```diff
@@ -83,24 +84,144 @@ crop_to_boundary <- function(boundary_shape_data, county_shape_data) {
...
+add_precinct_id <- function(precincts_long, county_name) {
+  precincts_long[, Precinct_I := paste0(county_name, "_", precinct_number)]
+  precincts_long[, precinct_number := NULL]
+  return(precincts_long)
+}
+
+# case-insensitively match each county-provided location name against the
+# state shapefile's USER_POLL_ naming. Name mismatches create a 
+# stop()ping error. Resolving requires either
+# editing the state source file or confirming a data with the county.
```
antisocialscientist, line 120: foo

antisocialscientist (approved):

PR #285, opened 2026-07-01
Precinct assignment and block validation

Amasus:
## Summary

Investigates and documents the precinct 73 population discrepancy (#283), where `assign_block_to_dominant_precinct()` incorrectly reports `Monongalia_73` as having zero population. The root cause is the winner-take-all dominant-block heuristic: when multiple small precincts compete for overlap with the same census blocks (as the six Skyview-cluster precincts do), no single block may ever have precinct 73 specifically as its single largest-overlap precinct, even though 73 has real area (~621,510 sq m) and is assigned to Skyview Elementary School.

## Changes

- Generalized precinct name/ID reconciliation logic (#282)
- Documented precinct 73 area and polling assignment facts
- Prepared findings for client clarification on whether the heuristic should be replaced with population-weighted-by-overlap or if zero-population precincts should be flagged for human review instead

## Notes

Blocks #282 (zero-population precinct drop logic) until the decision on how to handle precincts with real area but zero observed population is confirmed with the county/state.

**Review only** at this point. Do not close until we have feedback from client.


antisocialscientist (commented):
asdf

R/result_analysis/Extraction_configs/Monongalia_County_WV_no_precinct_data.r
```diff
@@ -0,0 +1,27 @@
...
+COUNTY_NAME <- sub("\\_.*", "", LOCATION)
+
+STATE_PRECINCT_SOURCE_FILE <- "datasets/precincts/West_Virginia_20260424_wmA84/VotingPrecincts_20260424_wmA84.shp"
+COUNTY_PRECINCT_SOURCE_FILE <- paste0("datasets/polling/", LOCATION, "/", LOCATION, "_potential_locations.csv")
+
+########
+# County-provided precinct/polling-location reconciliation (#282)
+########
+COUNTY_PROVIDES_PRECINCT_DATA <- FALSE
+COUNTY_PROVIDED_PRECINCT_FILE <- "temp/Precincts_by_Location.csv"
```
antisocialscientist: Is it better practice to leave this an empty string?

R/result_analysis/utility_functions/shape_extraction_functions.r
```diff
@@ -98,16 +98,121 @@ extract_county_precincts <- function(precinct_source_file, county_name,
   return(county_precincts)
 }
 
+##### reconcile a county-provided precinct/polling-location file
+##### against the state-extracted county_precincts #####
+
```
antisocialscientist, line 115: How is this different than the function I looked at last time?

R/result_analysis/utility_functions/shape_extraction_functions.r
```diff
@@ -135,6 +240,23 @@ assign_block_to_dominant_precinct <- function(county_precincts, county_blocks,
...
+# block.
+#NOTE: Assumes that each precinct is a union of blocks. Otherwise
+#the dominant precinct logic does not apply.
+assign_block_to_dominant_precinct <- function(county_precincts, county_blocks,
+                                          p3_population, area_crs = AREA_CRS) {
+
+  block_precinct_intersection <- compute_block_precinct_overlaps(
+    county_precincts, county_blocks, p3_population, area_crs
+  )
+
```
antisocialscientist, line 297: Is this here instead of earlier because you want to run the tests before assigning the blocks or is it something deeper?

R/result_analysis/utility_functions/shape_extraction_functions.r
```diff
@@ -147,6 +269,27 @@ assign_block_to_dominant_precinct <- function(county_precincts, county_blocks,
   return(block_precinct_intersection)
 }
 
+# unlike assign_block_to_dominant_precinct() (one row per block, its single
+# dominant precinct), this keeps every precinct a block significantly
+# overlaps: one row per (block, precinct) pair where the precinct holds more
+# than min_percent_overlap of the block's area. A block with significant
+# overlap in 3 precincts produces 3 rows. See #283.
```
antisocialscientist: Probably don't want to reference a ticket number inside the comments

R/result_analysis/extract_precincts.r
```diff
@@ -16,18 +16,18 @@ source("R/result_analysis/utility_functions/shape_extraction_functions.r")
...
-  config_path <- paste0("R/result_analysis/Extraction_configs/", args[1])
-  source(config_path)
-}
+#args <- commandArgs(trailingOnly = TRUE)
+#if (length(args) != 1) {
+#  stop("Must enter exactly one config file")
+#} else {
+#  config_path <- paste0("R/result_analysis/Extraction_configs/", args[1])
+#  source(config_path)
+#}
```
antisocialscientist, line 26: I can't figure out what is changed here

R/result_analysis/extract_precincts.r
```diff
@@ -16,18 +16,18 @@ source("R/result_analysis/utility_functions/shape_extraction_functions.r")
...
 
 ###
 # For inline testing only
 ###
-# source("R/result_analysis/Extraction_configs/Monongalia_County_WV.r")
+source("R/result_analysis/Extraction_configs/Monongalia_County_WV.r")
```
antisocialscientist, line 31: Again, what's the change?

R/result_analysis/extract_precincts.r
```diff
@@ -76,6 +85,59 @@ if (!file.exists(file.path(here(), precinct_analysis_output_folder))) {
...
+    address_col = COUNTY_PRECINCT_ADDRESS_COL,
+    county_precincts = county_precincts,
+    county_name = COUNTY_NAME
+  )
+} else {
+  # No county-provided data to reconcile against: fall back to dropping
+  # precincts with zero population, per block-level population data in
+  # block_precinct_assignment (#283 caveat: the dominant-block population
+  # heuristic can show real, populated precincts as zero -- e.g.
+  # Monongalia_73 -- so this fallback isn't fully trustworthy yet)
```
antisocialscientist: Is this still true?

antisocialscientist:
I have gone through the non-scratch files. If you want me to also look at scratch_* just let me know

antisocialscientist (commented):
I have resolved the comments that were a result of my being a doofus who can't read

R/result_analysis/utility_functions/shape_extraction_functions.r
Amasus: A function got broken into two because i wanted an interim result as an output

R/result_analysis/extract_precincts.r
Amasus: We are still waiting on confirmation from the client on this

Amasus:
@antisocialscientist -- summarizing what's changed since your last review (anchored at 1fed2051, the precinct-73 investigation commit). A lot landed since then; here's the digest rather than asking you to re-read every commit individually.

## Two accidental regressions, found and fixed

The branch picked up two `delivery/Monongalia_County` merges along the way, which caused real damage that took some digging to find:
- The first merge silently dropped the `flag_overlapping_blocks()` call from `extract_precincts.r` (both branches had independently rewritten the same region, so it merged cleanly with no conflict marker) and left two duplicate definitions of the reconciliation helpers with *reversed* error-message semantics (one told you to edit the state file on a mismatch, the other told you to edit the county file) -- fixed in `72fb9814`.
- A later commit (`dec1ffd4`) moved `precinct_analysis_output_folder`'s definition into `shape_extraction_functions.r` as top-level code, which depends on the config's `LOCATION` constant -- but that file is `source()`'d *before* the config in `extract_precincts.r`, so every run broke with `object 'LOCATION' not found`. Fixed in `f0bb7abd`, which also carried over some reconciliation-function renames (`check_poll_precinct_agreement`, `reconcile_state_precinct_data`) that had landed on the wrong branch during a session mix-up.

## The substantive change: stop silently dropping data, flag it instead

- **Zero-population precincts** are no longer silently filtered out of the resolved precinct list. A new step (Step 6) flags them -- covering both causes: dominant-assigned blocks summing to zero population, *and* a precinct never winning any block's dominant assignment at all (the actual #283 failure mode, e.g. `Monongalia_2B`). Both are distinguishable via `assigned_blocks`/`total_population` columns and written to `flagged_unpopulated_precincts.gpkg` with geometry intact, rather than dropped with no trace. `precincts_resolved` itself stays untouched/immutable throughout.
- **Both reconciliation hard gates** (`match_location_names()`, `check_poll_precinct_agreement()`) now write their mismatched rows to `location_mismatches.csv` / `location_precinct_mismatches.csv` before halting, instead of only surfacing them in the `stop()` message text. Design rationale posted to https://github.com/Voting-Rights-Code/Equitable-Polling-Locations/issues/283#issuecomment-4899356241.
- Step 5 (reconciliation) now runs right after state precincts are extracted (Step 2), before blocks are even loaded -- no reason to leave the two data sources unreconciled that late. Steps are renumbered 1-7 accordingly, no more duplicate/skipped numbers.
- `compute_block_precinct_overlaps()` (the shared `st_intersection()`) is computed once per run and passed into both `assign_block_to_dominant_precinct()` and `flag_overlapping_blocks()`, instead of being recomputed independently by each.

## Final cleanup

Removed `reconcile_location_precinct_pairs()` (dead code -- `reconcile_state_precinct_data()` had its own inlined equivalent) and `state_precincts_proj` (dead variable, never consumed in this file's history -- confirmed `compute_block_precinct_overlaps()` already does its own internal `st_transform()`, so no functionality was lost). Wired `flag_overlapping_blocks()`'s `min_percent_overlap` parameter into its actual threshold check (previously hardcoded 0.05/0.95, ignoring the parameter). Fixed a couple of typos and an unclear doc comment along the way.

## Verification

Ran `extract_precincts.r` end to end against real Monongalia data: all four outputs (`flagged_assigned_blocks.gpkg`, `flagged_overlapping_blocks.gpkg`, `flagged_unpopulated_precincts.gpkg`, `distance_flagged_blocks.csv`) regenerate correctly, including `Monongalia_73` and `Monongalia_2B` correctly flagged as unpopulated for their two distinct causes. Also exercised both reconciliation gates' mismatch-write paths against synthetic state/county data covering both mismatch directions.

Would appreciate a review with an eye toward merging into `delivery/Monongalia_County` when you get a chance.

Amasus:
Adding a comment to complete the documentation loop. 

The client responded to the communication [noted in 283](https://github.com/Voting-Rights-Code/Equitable-Polling-Locations/issues/283#issuecomment-4823671550) with 
> For Precinct 73, it would probably be best to proceed without assigning any census blocks to this precinct. The municipality did not update their boundaries after an annexation, so this information will not align correctly with the census blocks.

This communication lead to the decision to create the step that logs 0 population precincts.

Issue #264, opened 2026-06-18
CVAP distance/source files are named by census decade, not actual RDH vintage — breaks traceability


Amasus:
## Problem

Since CVAP pulling was changed to fetch the most recent RDH release within a census decade (rather than always the original decennial year), the actual vintage of CVAP data used is lost on disk. Every file involved is named only by the census decade (e.g. `2020`), never by the actual RDH release year (e.g. `2023`).

## Where this happens

1. [`pull_state_CVAP_data`](python/utils/pull_census_data.py#L495) computes `most_recent_year` to select which RDH dataset to download, but only returns the `DataFrame` — the vintage year is discarded after selection.
2. [`pull_CVAP_data`](python/utils/pull_census_data.py#L582) calls `save_CVAP_data(locality_CVAP, census_year, county_ST)` using the original decade marker, not the actual vintage pulled.
3. [`build_CVAP_source_file_path`](python/utils/utils.py#L195) names the saved file `CVAP_{census_year}-Data.csv` — so the raw CVAP CSV on disk is indistinguishable between a 2020 pull and a 2023 pull.
4. [`get_CVAP_demographics`](python/solver/model_data.py#L286) reads that file back using the same decade-keyed path.
5. [`build_distance_file_path`](python/utils/utils.py#L106) inherits the same `census_year` for the distance data file name, so the loss compounds into the final intermediate dataset filename too.

**Impact:** Given only a distance data file (or the saved CVAP CSV), there is no way to determine which actual RDH CVAP release backs the data — breaking traceability/reproducibility for CVAP-based runs.

## Options to discuss

- **Minimal:** Have `pull_state_CVAP_data` also return the vintage year; embed it in the CVAP filename and glob for it (similar to how TIGER shapefiles are already looked up by globbing rather than a fixed path).
- **Also fixes distance file:** Thread the vintage year into `build_distance_file_path` so the distance CSV name reflects the actual CVAP vintage used, not just the census decade.
- **More invasive:** If CVAP results are also written to the database, `distance_data_sets` (or related) metadata may need a new column to record CVAP vintage for DB-backed runs.

## Next step

Needs a decision on where the vintage year should be recorded (filename vs. separate metadata) before implementation, since it affects existing naming conventions and lookup logic for CVAP files.

Issue #246, opened 2026-06-18
[RDH-pop] Solver layer: get_RDH_population_demographics() and dispatch


Amasus:
Part of #243. Blocked on #244.

Parallel to `get_CVAP_demographics()` in `model_data.py`. Add a new `census_data_type` value (name TBD based on #244) to the dispatch in `get_demographics_block()`. Column rename mapping should produce the standard `DISTANCE_*` columns the rest of the solver expects.

**Acceptance check**: verify that `build_distance_data()` correctly calls `get_demographics_block()` and automatically writes a distance dataset against the testing fixture data (from #245/#244) — no separate manual creation step should be needed, parallel to how `testing_CVAP_distances_2020.csv` works. Full real-county verification happens at the milestone level (see #243).

Issue #245, opened 2026-06-18
[RDH-pop] Download layer: pull_RDH_population_data()


Amasus:
Part of #243. Blocked on #244.

Parallel to `pull_CVAP_data()` / `pull_state_CVAP_data()`. Implement the catalog query, county filter, and save-to-disk logic. Reuses existing RDH credentials. Exact field names and save path depend on #244 findings.

**Directory structure**: add a new folder under `datasets/census/`, named after the new `census_data_type` string (parallel to `datasets/census/CVAP/` and `datasets/census/redistricting/`), with a per-location subfolder structure matching CVAP's. Add the corresponding `<NEW_TYPE>_FOLDER_NAME` constant in `directory_constants.py` and `build_<NEW_TYPE>_dir_path()` / `build_<NEW_TYPE>_source_file_path()` functions in `utils.py`, parallel to `build_CVAP_dir_path()` / `build_CVAP_source_file_path()`.

DashaU3:
Issue with downloading data: What works (CVAP, redistricting data, etc.): RDH exposes these via a catalog API (wp-json/download/list) that we can query programmatically with username/password. We search the catalog, get a file URL, and download it — fully automated, no browser required.

What doesn't work (VAP projections 2026–2035): These files are not indexed in the catalog API at all. The only access route is a browser download page (redistrictingdatahub.org/download/?datasetid=...&document=...) that requires an active browser login session (WordPress session cookie). We confirmed that:

The catalog API returns zero results for these files regardless of search parameters
The download endpoint returns an HTML login page for every programmatic request we tried (query params, POST body, HTTP Basic Auth, HTTP session)
WordPress's login endpoint (wp-login.php) returns 403 from this environment, so we can't programmatically obtain a session cookie either
Consequence: Every new state whose VAP projection data you need requires a manual browser download (log in to RDH, click download, unzip, place the CSV in the right directory). Each state file covers all counties in that state, but you'd need a separate download per state.

Potential paths forward to discuss:

Ask RDH to add VAP projection files to their catalog API — this would make it fully automated like CVAP
Pre-download all needed state files manually once and commit them (or store them in a shared location) so the team doesn't hit this during model runs
Accept the manual step and document it clearly in docs/input_files.md

PR #286, opened 2026-07-01
feat(utils): add RDH_population directory structure and path builders

DashaU3:


Amasus:
Just a few good habits to get into:
1. Link the actual ticket to the PR, so that the reviewer has an easy time understanding what they are reviewing. This can go in the description.
2. Give a brief description of the work you did and why in the PR as well

python/utils/directory_constants.py
```diff
@@ -24,6 +24,8 @@
 
 CVAP_FOLDER_NAME = 'CVAP'
 
+RDH_POPULATION_FOLDER_NAME = 'RDH_population'
```
Amasus: Can we name this something more descriptive. Like Predicted_VAP? or RDH_predicted_VAP? Both for the variable, the string and the associated functions, .csvs etc.

Otherwise, it is confusing what this data is.

python/utils/utils.py
Amasus, line 1: favor. Looking at the code directly, I see that there is a floating return on line 183(ish) of utils.py. Can you just delete that? Chad and I both missed it.

python/tests/pull_census_data_test.py
```diff
@@ -420,6 +420,30 @@ def test_pull_CVAP_data_raises_when_rdh_credentials_missing(monkeypatch):
         pcd.pull_CVAP_data('GA', 'Gwinnett County', '2020', census_apikey='fake-key')
 
 
+class TestBuildRDHPopulationPaths:
```
Amasus, line 423: Claude sometimes takes "test functions" too literally. There isn't a similar testing done for either the census or the CVAP folders. This feels like a bit of over kill to me. I believe that if these functions don't work properly, other things will break in the future and it will be caught easily. However, if you (or Claude) want to push back why ALL the tiny little build directory structures need to be tested, I'm open to listening. But it really should be all or none.

python/utils/utils.py
```diff
@@ -201,6 +201,32 @@ def build_CVAP_source_file_path(census_year: str, location: str) -> str:
...
+    Args:
+        census_year: Decennial census year string, e.g. '2020'.
+        location: Location identifier, e.g. 'Gwinnett_County_GA'.
+
+    Returns:
+        Absolute path to the RDH_population CSV file.
+    """
+    return os.path.join(
+        build_RDH_population_dir_path(location),
+        f'RDH_population_{census_year}-Data.csv',
```
Amasus: 1. Have you confirmed that this is the naming structure for the RDH file being downloaded?

2. Presumably, when we are predicting a population for 2026, based off the 2020 census, we will have to do a different pull from the dataset than if we were pulling 2024 data. Therefore, the prediction year needs to be recorded somewhere as well. Whether this happens at the file level or the folder level is an interesting question. Talk to Chad about his thoughts on issue #264 (the one about data vintage and CVAP).

Amasus (changes requested):
Take a look at the comments. If you have questions, just ask

Amasus:
The ticket here (#245) talks about building pull RDH data, but the work here only touches the utils. 
The next ticket(#246) has a stub for pull RDH data. When is that going to get built?

PR #290, opened 2026-07-03
feat(utils): add RDH predicted whole population path builders

DashaU3:
## Summary

- Adds `RDH_PREDICTED_WHOLE_POPULATION_FOLDER_NAME` constant to `directory_constants.py`
- Adds `build_RDH_predicted_whole_population_dir_path(location)` to `utils.py` — returns `datasets/census/RDH_predicted_whole_population/<location>/`
- Adds `build_RDH_predicted_whole_population_source_file_path(location, projection_year)` to `utils.py` — scans the directory for any `.csv` rather than constructing a fixed name, since RDH filenames encode state abbreviation and year range (e.g. `ga_pop_proj_2026_2035_b.csv`) and cannot be derived from the location string alone; falls back to a synthetic path for a meaningful error message when no file is present yet

Follows the existing pattern used for `CVAP` and `redistricting` path builders.

## Test plan

- [ ] `pytest python/` passes

### DIFF
--- a/python/utils/directory_constants.py
+++ b/python/utils/directory_constants.py
@@ -24,7 +24,7 @@ REDISTRICTING_FOLDER_NAME = 'redistricting'
 
 CVAP_FOLDER_NAME = 'CVAP'
 
-RDH_PREDICTED_WHOLE_POPULATION_FOLDER_NAME = 'RDH_predicted_whole_population'
+RDH_PREDICTED_VAP_FOLDER_NAME = 'RDH_predicted_vap'
 
 BLOCK_GEO = 'block'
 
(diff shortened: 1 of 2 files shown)


python/utils/directory_constants.py
```diff
@@ -24,6 +24,8 @@
 
 CVAP_FOLDER_NAME = 'CVAP'
 
+RDH_PREDICTED_WHOLE_POPULATION_FOLDER_NAME = 'RDH_predicted_whole_population'
```
Amasus: should be for VAP, not all population

antisocialscientist:
I think that this is clear and works. I think your documentation of the new functions in  utils.py have very verbose comments, can you make them match the rest of the utils.py in terms of comment verbosity?

Amasus (approved):

PR #291, opened 2026-07-03
feat(solver,utils): add predicted_vap census data type

DashaU3:
Stacks on #290.

## Summary

- **`model_data.py`**: `RDH_PREDICTED_VAP_COL_PREFIXES` constant with correct column names confirmed against real RDH data; `get_RDH_predicted_vap_demographics(location, projection_year)`; `projection_year` promoted to required parameter in `get_demographics_block` and `build_distance_data`
- **`pull_census_data.py`**: `pull_RDH_predicted_vap_data` raises `NotImplementedError` with manual download instructions and per-state dataset IDs (`VAP_PROJ_BLOCK_DATASET_IDS`); `locality_predicted_vap_only`, `save_RDH_predicted_vap_data`; `RDH_DOWNLOAD_URL` constant
- **`pull_rdh_vap_playwright.py`** (new): Playwright-based automated download — logs into RDH at `/login/`, navigates to the per-state confirmation page, extracts the pre-signed S3 URL from the "Download" anchor (rendered via JS), fetches with `requests`. Confirmed working for GA / Gwinnett County.
- **`pull_rdh_vap_cli.py`** (new): Thin CLI wrapper around `pull_RDH_predicted_vap_data`
- **`model_data_test.py`**: `DemographicsTestBase` eliminates duplicated `test_output_columns`, `test_geoid_is_string_dtype`, and `test_missing_directory_triggers_pull` across all three demographics test classes
- Threads `projection_year` through `model_run.py`, `conftest.py`, and `db_import_distance_data_cli.py`
- Replaces synthetic testing fixture with 296 real Gwinnett County census blocks from `ga_vap_proj_2026_2035_b.csv` (3 tracts); un-ignores `datasets/census/RDH_predicted_vap/testing/*`

### Column name fix

Column prefixes were corrected after downloading real RDH data. The actual CSV uses:
- `Projected_WhiteAloneNotHisp_VAP` (not `Projected_NotHisp_WhiteAlone_VAP`)
- `Projected_BlackOrAfAmAlone_VAP` (no `NotHisp_` prefix)
- Same pattern for AIAN, Asian, Pacific Islander, Other, Two+ races

## Test plan

- [ ] `pytest python/` passes
- [ ] `pylint python/` passes (no new warnings)


python/scripts/db_import_distance_data_cli.py
```diff
@@ -55,11 +55,12 @@ def import_distance_data(
 def build_and_import_distance_data(
     query: Query,
     census_year: str,
-    census_data_type:str, 
+    census_data_type: str,
     location: str,
     driving: bool,
     maps_source_date: str,
     log_distance: bool,
+    projection_year: str = None,
```
Amasus: should this be str instead of str = None

python/utils/pull_census_data.py
```diff
@@ -634,6 +635,167 @@ def pull_CVAP_data(
...
+        rdh_url: RDH list endpoint. Defaults to RDH_LIST_URL.
+
+    Returns:
+        The string 'Success' on completion.
+
+    Raises:
+        ValueError: If credentials are missing, the county is not found in the state data,
+            or no block-level projection file exists in the RDH catalog.
+    '''
+    if state_lookup is None:
```
Amasus: Why this logic instead of having just kwargs in the function call?

python/solver/model_config.py
```diff
@@ -124,6 +124,9 @@ class PollingModelConfig:
     map_source_date: str = None
     ''' The date (YYYYMMDD) of the maps source to use if driving distances are used. '''
 
+    projection_year: str = None
```
Amasus: Two things: 
1. To keep this human readable, the config fields are currently organized: things that go in a yaml file, things that are needed for the database. Since the projection year goes in the yaml file, can it be grouped with the rest of them?
2. Also, setting this to str = None makes it an optional field. There's nothing else in the yaml fields that show up as optional. I don't think this is a good idea. Though I'd defer to @abd1tus  for better logic. 

Can you find time to investigate this with him?

python/solver/model_data.py
```diff
@@ -330,7 +346,76 @@ def get_CVAP_demographics(census_year: str, location: str):
...
+    # for the file — this mirrors the pull-on-demand pattern used by CVAP and redistricting.
+    if not os.path.exists(vap_dir):
+        statecode = location[-2:]
+        locality = location[:-3].replace('_', ' ')
+        pull_RDH_predicted_vap_data(statecode, locality, '2020')
+
+    if not os.path.exists(vap_source_file):
+        raise ValueError(
+            f'RDH predicted VAP data not found. Download using '
+            f'pull_RDH_predicted_vap_data or manually following the instructions in '
```
Amasus, line 374: 1. Do you mean pull_RDH_predicted_vap_playwright()? Why does this need to be a manual step and not automated?

python/solver/model_data.py
```diff
@@ -330,7 +346,76 @@ def get_CVAP_demographics(census_year: str, location: str):
...
+
+    # Validate that the requested projection year is actually present in the file.
+    missing_cols = [col for col in vap_projection_year_cols if col not in vap_df.columns]
+    if missing_cols:
+        raise ValueError(
+            f'projection_year {projection_year!r} not found in RDH VAP file for {location}. '
+            f'Missing columns: {missing_cols}'
+        )
+
+    vap_df = vap_df[['GEOID'] + vap_projection_year_cols].copy()
```
Amasus: As Chad said, this is a constant somewhere already. Make use of that.

python/scripts/db_import_distance_data_cli.py
Amasus: This is str | None later. Should it be that here as well?

python/tests/model_data_test.py
```diff
@@ -15,7 +15,7 @@
     DISTANCE_ID_ORIG, DISTANCE_ID_DEST, DISTANCE_ADDRESS,
     DISTANCE_DEST_LAT, DISTANCE_DEST_LON, DISTANCE_ORIG_LAT, DISTANCE_ORIG_LON,
     DISTANCE_LOCATION_TYPE, DISTANCE_DEST_TYPE,
-    DISTANCE_OTHER, DEMOGRAPHICS_OUTPUT_COLUMNS, DISTANCE_DEMOGRAPHICS_COLUMNS,
+    DISTANCE_OTHER, DISTANCE_TOTAL_POPULATION, DEMOGRAPHICS_OUTPUT_COLUMNS, DISTANCE_DEMOGRAPHICS_COLUMNS,
```
Amasus: distance_total_population is in distance_demographics_columns already

python/tests/model_data_test.py
```diff
@@ -189,56 +190,98 @@ def test_clean_data(testing_config_driving, location_df_with_driving):
         )
 
 
-class TestGetCVAPDemographics:
-    ''' Unit tests for model_data.get_CVAP_demographics using the committed testing fixture. '''
+class DemographicsTestBase:
```
Amasus, line 193: What is a patch_target? This just needs a line (not a Claude-novel) comment explaining

python/utils/pull_census_data.py
```diff
@@ -634,6 +652,166 @@ def pull_CVAP_data(
...
+    if rdh_username is None or rdh_password is None:
+        loaded_username, loaded_password = _load_rdh_credentials()
+        rdh_username = rdh_username or loaded_username
+        rdh_password = rdh_password or loaded_password
+    if rdh_username is None or rdh_password is None:
+        raise ValueError(
+            'No RDH credentials available. Run `python run.py secret set rdh`. See README.'
+        )
+
+    # --- FIPS code lookup ---
```
Amasus, line 767: The FIPS code lookup is now repeated code in 3 places. Can this be made into a function and inserted in all three?

python/utils/pull_census_data.py
```diff
@@ -634,6 +652,166 @@ def pull_CVAP_data(
...
+
+    # The /download/ endpoint requires session-based auth (browser cookies), not query-param
+    # auth like the wp-json API. Log in first to get the WordPress session cookies.
+    # RDH's /download/ endpoint requires a browser session cookie (wp-login.php is
+    # blocked for programmatic access). These files must be downloaded manually.
+    # Visit the URL below in a browser while logged in to redistrictingdatahub.org,
+    # then save the CSV from the zip to:
+    #   datasets/census/RDH_predicted_vap/{county_st}/
+    raise NotImplementedError(
+        f'Automated download of VAP projection data is not supported: the RDH download '
```
Amasus: This is killing the function. What's the actual solution here? Should this function be here at all if it is going to die like this? Can this be replaced by the playwright option?

python/scripts/pull_rdh_vap_cli.py
```diff
@@ -0,0 +1,27 @@
+"""Command line utility to pull RDH predicted VAP data for a given state and county."""
```
Amasus: This function won't work at all, correct? given that pull_RDH_predicted_vap_data doesn't work?

python/scripts/pull_rdh_vap_playwright.py
```diff
@@ -0,0 +1,133 @@
+"""
+Experimental: download RDH VAP projection data using Playwright browser automation.
```
Amasus: This should be refactored into pull_census_data so that the RDH pull happens automatically. 
I don't think there needs to be a _cli version of this pull function.

Amasus (changes requested):

Amasus (changes requested):
can you fix the merge conflicts here and I'll merge it in?

Issue #270, opened 2026-06-20
[Precincts] Produce county-level heat map of blocks far from their polling location


Amasus:
Part of milestone "Monongalia County precinct extraction & distance flagging". Depends on #269.

County-level plot of all Monongalia blocks (`ggplot()` + `geom_sf()` + `ggsave()`, following the visualization pattern already in `extract_city.r`):

- Blocks under 5 miles from their assigned polling location: neutral/gray fill.
- Blocks at or over 5 miles: red fill, graduated by distance (farther = brighter red).
- Precinct boundaries drawn on top as outlines for context.

PR #293, opened 2026-07-04
feat(precincts): add county-level distance heat maps (#270)

Amasus:
## Summary
- Adds `make_demo_distance_heat_map()` to `shape_extraction_functions.r`, wired into `extract_precincts.r` as Step 7, producing county-level maps of blocks by driving distance to their assigned polling location.
- Two modes: a choropleth (red graduated by distance for flagged blocks) and a dot mode (mirrors `make_demo_dist_map()` in `map_functions.R` — a dot per flagged block, sized by a demographic's population and colored by distance).
- Gray = no population, white = populated (under threshold in choropleth mode, or context fill in dot mode), so the distance encoding stands out.
- Legend/title labels go through a new `demo_pop_legend_dict` (mirrors `graph_functions.R`'s `demographic_legend_dict`, extended for this pipeline's column names) for consistent naming (e.g. `black` → "African American").
- Maps are reprojected to `CRS_PROJECTION` before plotting so the graticule renders horizontal/vertical, matching the project's other maps.

Builds on #269, which merged into `delivery/Monongalia_County` via #292 (this diff is scoped to just the heat-map work: 7 files, +453/-49).

## Test plan
- [x] Ran `R/result_analysis/extract_precincts.r` end-to-end against real Monongalia County data (choropleth mode, dot mode with `total_population`, dot mode with `black`)
- [x] Visually inspected all three output PNGs for correct fills, legends, titles, and captions
- [x] `lintr::lint()` on both changed files — no new warnings beyond pre-existing false positives (NSE column references, global-default-arg pattern already used elsewhere in the file)


R/result_analysis/Extraction_configs/Monongalia_County_WV_id_mismatch.r
antisocialscientist, line 37: Just confirming that it is okay given that there can be missing values here, just not missing columns

antisocialscientist: This correctly explains what you are doing, but do you want them ALL to have to be null or just AT LEAST ONE?

If the latter, should the any() be an all()

R/result_analysis/utility_functions/shape_extraction_functions.r
antisocialscientist, line 358: I think this is correct, and the column name is Location (with the initial capitalization) and not location (all lower) Can you just confirm?

antisocialscientist:
Very close to ready to go. I made comments on one of the configs that apply to them all. I also am leaving the reference to the still open issue on replacing the 5 mile distance with actual time since it is still outstanding.

R/result_analysis/Extraction_configs/Monongalia_County_WV_id_mismatch.r
Amasus: correct. It's not so much about the column being missing as, if the county gives us data, which column should we be looking at for the precinct names

Amasus: it should throw an error if any are NOT null. I'll fix the error message

R/result_analysis/utility_functions/shape_extraction_functions.r
Amasus: This code runs. It wouldn't if the capitalization was wrong. Do you need me to look this up still?

Issue #294, opened 2026-07-04
driving_time: county-scale OOM (config hardness) + log_distance broken for minutes — findings & options


abd1tus:
## Summary

While validating the `driving_time` metric end-to-end on **Tarrant County TX** (full county, census-block-level origins → an ~8.5M binary-variable MILP), we hit two independent problems and fixed a third. None of these block the remaining epic phases (#230 BigQuery `duration_min`, #231 R toolkit) — they concern actually *running* large `driving_time` optimizations. Filing so we can resume once Susama provides real-world driving-time configs to test against.

Part of #227. Related: #226 (ORS cross-border), #288, and the seconds→minutes change (commit `57df3201`).

---

## Finding 1 — county-scale OOM is caused by the *optimization parameters*, not the metric

A `driving_time` Tarrant run (`precincts_open: 215`, `minpctold: 0.5`, `year: ['2024','2025']`) was **SIGKILL'd (`-9`, OOM)** by SCIP while building the root LP, after presolve, needing **> 48 GB RAM + 8.7 GB swap**. The model: **8,534,656 binary vars / 8,557,849 constraints**.

We isolated the cause with controlled runs (~48 GB dev container, SCIP 10.0.2):

| Config | Params | Presolve result | Outcome |
|---|---|---|---|
| `driving_time` (as written) | `precincts_open: 215`, `minpctold: 0.5` | deletes **1** constraint → 8,557,848 left | **OOM** building root LP |
| haversine `year_2024` | `precincts_open: null`, `minpctold: 1`, `maxpctnew: 0` | deletes **8,534,309** conss → **23,540** left | solves at node 1 (gap 1.87%, ~53 G) |
| `driving_time` w/ easy params | `precincts_open: null`, `minpctold: 1` | deletes **440,997 vars / 8,534,309 conss** (bit-identical to haversine) → 23,540 | **no OOM** |

**Conclusion:** the model's *variable/constraint count is identical across metrics* (assignment vars = origins × candidate-locations). When site selection is effectively fixed (`minpctold: 1, maxpctnew: 0`), presolve collapses the ~8.5M `setppc` constraints ~363× and it solves. When site selection is free (`precincts_open: 215, minpctold: 0.5`), presolve can't reduce it and the full LP OOMs. **The metric, the 6.36 GB distance file, and data size are innocent** — the same hard parameters in haversine would OOM too.

**Possible solutions / notes:**
[...]

Issue #227, opened 2026-06-01
Add driving time (duration) as an alternate metric to driving distance


abd1tus:
## Context

Today the toolset routes every census-block-centroid → polling-place pair through ORS and records **driving distance in meters** (`distance_m`); the Kolm-Pollak solver then minimizes a function of that distance and penalizes its inequality. This ticket proposes adding **driving time (duration)** as an alternate metric the pipeline can produce and the model can optimize against.

**Why it's worth having.** For equity, *time* is often a truer measure of access burden than raw distance: a 10 km drive on a rural 2-lane road is a heavier burden than 10 km of highway, and time captures that where distance does not. Letting an analysis minimize/equalize **travel time** rather than **travel distance** is a meaningful, arguably more representative objective for some jurisdictions. The KP objective is mathematically unit-agnostic (it operates on a per-person metric value and an inequality penalty), so feeding seconds instead of meters yields a coherent "minimize travel time, penalize time inequity" optimization.

**This is low-risk on the routing side and mostly a plumbing/schema question downstream.** ORS already returns durations.

## What's already in place

- `python/utils/ors_client.py::query_matrix` already takes `metric='distance' | 'duration'` and sets `metrics: [metric]` on the ORS request (covered by `ors_client_test.py:62`). ORS `/v2/matrix` can even return **both** metrics in one call (`metrics: ['distance', 'duration']`).
- `query_directions` (the per-pair retry path) currently parses only `segments[0].distance`; ORS also returns `segments[0].duration`, so the duration is available there too with a small change.

## Touchpoints (everything downstream is hardcoded to distance)

- **`python/solver/constants.py`** — only `DISTANCE_DISTANCE_M = 'distance_m'` exists; no duration constant. Add e.g. `DISTANCE_DURATION_S = 'duration_s'`, plus source-label constants (cf. `DISTANCE_SOURCE_DRIVING_DISTANCE = 'driving distance'`).
- **`python/utils/driving_distance_matrix.py`** — `build_distance_matrix` / `matrix_response_to_long_df` assume a single `distance_m` column. `estimate_origin` snaps unroutable origins using **haversine distance in meters** — that snap doesn't translate to seconds without a speed assumption (see design Q3).
[...]

PR #296, opened 2026-07-08
fix(solver): validate distance/time non-negativity before the log transform

### LINKED ISSUE #295: Validate distance/time non-negativity before the log transform, not after
## Problem

`filter_distance_data`'s non-negativity guard runs **after** the log transform, so on a `log_distance` run it validates log-space values instead of raw meters/seconds.

Pipeline: `build_distance_data` computes raw distances, applies `_log_transform_metric_columns` (`model_data.py:465`) when `log_distance` is set, and **caches the log-transformed values** to the `..._log.csv` distance file. `filter_distance_data` runs later, at model construction in `run_setup` (`model_run.py:195` for the base set, `:198` for the alpha set), reading that cached frame — which for a log run is already in log space. The `distance_m < 0` guard (added in `4217f2e0`, "reject negative driving distances at ingestion") therefore checks log values.

Consequence: a legitimately small raw value produces a **negative log**, which the guard falsely rejects:

- `duration_s` in `(0, 1)` or `distance_m` in `(0, 1)` → `log < 0` → `ValueError: Some distances are missing or invalid (negative)`.
- The `0 → 0.001` floor in `_log_transform_metric_columns` makes an exactly-coincident pair `log(0.001) = −6.9`, likewise rejected.

This is the root of #294's `log_distance: true` crash for `driving_time` (e.g. a census block whose centroid sits on its polling site → ~0 driving time), and it affects any log run that contains a 0 / sub-1 raw value.

## Decision (per Susama)

- We should never have negative **raw** distances or times, so the non-negativity check is correct — but it **must run on raw values, before the log is taken**. A "log-meter" is not a meter; a negative log is a legitimately-small distance, not invalid data.
- Keep a small epsilon floor purely to avoid `log(0)`: `0 → 0.001` m (distance), `0 → 0.5` s (time). **Do not** floor to `1.0` — that masks bad data and would move existing baselines.

## Changes

1. `_log_transform_metric_columns`: floor *exactly-0* per metric — `distance_m → 0.001`, `duration_s → 0.5`. (Reverts the provisional `< 1 → 1.0` floor from `5d8fe5ba`.)
2. Move the non-negativity validity check to raw values in `build_distance_data`, **before** the log transform.
3. `filter_distance_data`: drop the `distance_m < 0` condition; keep the null/missing check (that one is per-config and correct post-log).

## Effect

Raw negatives are still rejected as true invalids; legitimate tiny distances log to finite negatives and pass; distance-log baselines revert to byte-identical with `main` (the `0.001` floor is restored).

Relates to #294, epic #227. Work off `feature/driving-time-metric`; PR back to the epic for review.

abd1tus:
Per Susama's call on #294 / #295: the distance/time non-negativity check must run on **raw** values, before the log transform — not after.

## Problem

The `distance_m < 0` guard (added in `4217f2e0`) lives in `filter_distance_data`, which runs **downstream** of the log transform: `build_distance_data` caches log-transformed distances to the `..._log.csv` file, and `filter_distance_data` reads that cached frame at model setup (`run_setup`, `model_run.py:195`/`:198`). So on a `log_distance` run it validates log-space values and falsely rejects legitimately small distances/durations whose log is negative. This crashed `driving_time` + `log_distance: true` on coincident block↔site pairs (#294), and would crash any log run with a 0 / sub-1 raw value.

## Changes

- **New `_reject_negative_metric_values`** in `build_distance_data` validates raw `distance_m` / `duration_s` non-negativity **before** the log ("at ingestion").
- **`filter_distance_data`** drops the `distance_m < 0` condition; keeps the null/missing check (correct post-log and per-config).
- **`_log_transform_metric_columns`** floors *exactly-0* per metric (`distance_m → 0.001` m, `duration_s → 0.5` s) purely to avoid `log(0)`; sub-1 positive values keep their finite negative logs. This replaces the provisional `<1 → 1.0` floor from `5d8fe5ba`.

## Effect

Raw negatives are still rejected as true invalids; legitimately small distances log to finite negatives and pass; **distance-log baselines revert to byte-identical with `main`** (the `0.001` floor is restored).

## Verification

TDD (red → green). `pytest -m "not e2e_db"`: 384 passed / 1 skipped. `model_data.py` lint 10.00/10.

Closes #295. Context: #294, epic #227.

### DIFF
--- a/python/solver/model_data.py
+++ b/python/solver/model_data.py
@@ -321,8 +321,9 @@ class BuildDistanceMetaData:
 # log, purely to avoid log(0) = -inf. Per-metric because a meter and a second are
 # different units; a coincident pair (a block centroid on its own site) is
 # essentially zero travel, so these keep the log finite and strongly negative.
-# Non-negativity of the RAW values is validated separately, before the log, by
-# _reject_negative_metric_values in build_distance_data.
+# Non-negativity of the RAW values is guaranteed upstream, when the driving matrix
+# is built (_coerce_negative_metrics_to_null in driving_distance_matrix), so this
+# floor only handles legitimate exactly-zero (coincident) pairs.
 LOG_ZERO_DISTANCE_FLOOR_M = 0.001
 LOG_ZERO_DURATION_FLOOR_S = 0.5
 
@@ -351,33 +352,6 @@ def _log_transform_metric_columns(distance_df: pd.DataFrame) -> pd.DataFrame:
         distance_df[column].mask(distance_df[column] == 0.0, zero_floor, inplace=True)
         distance_df[column] = np.log(distance_df[column])
     return distance_df
-
-
-def _reject_negative_metric_values(distance_df: pd.DataFrame) -> None:
-    '''Raise if any raw distance or duration is negative (before any log transform).
-
-    A negative meter or second is never a real-world value, and because the KP
-    objective minimizes distance the solver would treat such a pair as more
-    attractive than a genuine zero-distance pair. Validity is enforced here, on
-    the RAW values, rather than in filter_distance_data: that gate runs downstream
-    of the log transform, where a negative distance_m is log(sub-1) -- a
-    legitimately small distance, not invalid data. See issue #295.
-
-    Args:
-        distance_df: Frame holding raw distance_m and optionally duration_s.
-
-    Raises:
-        ValueError: If distance_m or duration_s contains a negative value.
-    '''
-    metric_columns = [DISTANCE_DISTANCE_M]
-    if DISTANCE_DURATION_S in distance_df.columns:
-        metric_columns.append(DISTANCE_DURATION_S)
-    for column in metric_columns:
-        if (distance_df[column] < 0).any():
-            raise ValueError(
-                f'{column} contains negative values, which are never valid raw '
-                f'distances or durations; check the distance/duration source data.'
-            )
 
 
 # Old Build source function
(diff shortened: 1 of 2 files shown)


python/solver/model_data.py
```diff
@@ -336,13 +344,40 @@ def _log_transform_metric_columns(distance_df: pd.DataFrame) -> pd.DataFrame:
...
+        ValueError: If distance_m or duration_s contains a negative value.
+    '''
     metric_columns = [DISTANCE_DISTANCE_M]
     if DISTANCE_DURATION_S in distance_df.columns:
         metric_columns.append(DISTANCE_DURATION_S)
     for column in metric_columns:
-        distance_df[column].mask(distance_df[column] < 1.0, 1.0, inplace=True)
-        distance_df[column] = np.log(distance_df[column])
-    return distance_df
+        if (distance_df[column] < 0).any():
```
Amasus: I still don't understand why this check belongs here and not when you are creating the driving_distance / driving_time data

abd1tus:
Addressed your review (commit e17829fe): removed the build-side `_reject_negative_metric_values` and its four tests. As you noted, negatives are already guarded when the driving data is created — `driving_distance_matrix._coerce_negative_metrics_to_null` replaces any negative ORS value with NaN (then snapped or dropped) before the CSV is written — so the build-side check was redundant. What remains is only the per-metric log floor for legitimate exactly-zero coincident pairs, plus the null-only check in `filter_distance_data`. Generation still coerces silently rather than raising; leaving that coerce-vs-raise choice for our discussion.

python/solver/model_data.py
abd1tus: The check as been removed in the latest changes  (commit e17829fe).  Negatives are handled at data creation.

Amasus (approved):

Issue #298, opened 2026-07-10
Heat maps (feature/make_distance_heat_maps) need to account for #277's duplicate/dropped precinct geometry


Amasus:
## Summary

Blocked by #297.

`feature/make_distance_heat_maps` (`R/result_analysis/utility_functions/map_functions.R`, not yet merged) visualizes solver results and block-group demographics using precinct/block assignment produced by `extract_precincts.r`. The #277 state-file versioning work introduced two behavior changes upstream of that map code that need to be accounted for before this branch merges:
1. Unassigned rows (A/B dropped) — make_precinct_map() already has a designed fallback for this: blocks with no id_dest (unpopulated/unmatched) get folded into their nearest assigned neighbor via st_join(..., join = st_nearest_feature), purely for visual continuity. That's pre-existing behavior, not something we broke. But dropping A/B doesn't just affect drawing — it changes the actual dominant-assignment step upstream (assign_block_to_dominant_precinct): any census blocks that used to compete for A/B's territory now get dominant-assigned to whichever real neighboring precinct has the next-largest overlap. Since A/B are zero-population, this shouldn't move any population-weighted numbers, but it could visibly shift a precinct boundary line in that specific area on the map — worth a visual spot-check once maps run for Monongalia, not just a numeric check.
2. Duplicate geometry (2A/2B) — make_precinct_map() groups by id_dest (the destination polling place), not by Precinct_I at all. Since 2A and 2B route to the same destination (Morgantown High School), they'd naturally union into one shape regardless — so the map itself should be robust to this. The real exposure is upstream: with with_ties = TRUE, a tied block now produces two rows (one for 2A, one for 2B) before grouping-by-destination happens. bg_result_geom()/demographically_weighted_distances() sum population before any dedup — so block-group population sums, and therefore dot sizes in make_demo_dist_map() and color intensity in make_bg_maps(), would be inflated for that shared area. You already said the double-count is acceptable for downstream KP-metric usage, but that same inflation now also shows up visually in the heat maps themselves, not just in aggregate stats.

3. Dropped precinct geometries — same mechanism as #1: assign_block_to_dominant_precinct no longer has Monongalia_2 as a candidate, only 2A/2B (its exact duplicates), so no net geometry is lost there — precinct 2's area doesn't disappear, it's just now claimed by 2A and 2B jointly.

## Before starting work here

Re-evaluate the status of the above changes and how they affect heat maps after #297 has been completed before starting this work. 
[...]

Issue #297, opened 2026-07-10
Precinct flagging/logging (#285) runs against corrected state file, not as-provided original


Amasus:
## Summary

The #277 state-file versioning work (this branch, `feature/state_data_governance`) introduces a stable path (`STATE_PRECINCT_SOURCE_FILE`) that always points at the *latest corrected* version of the WV statewide precinct shapefile, not the true as-provided original. That's intentional for the reconciliation gates themselves (`match_location_names()`/`check_poll_precinct_agreement()` need to check against current state to know what's still wrong). But it means every other step in `extract_precincts.r` that reads `state_precincts`/`precincts_resolved` -- all downstream of Step 2 -- now also runs against progressively-cleaned data, including the flagging/logging work done on #285 / #283:

- `flagged_assigned_blocks.gpkg` (Step 4)
- `flagged_overlapping_blocks.gpkg` (Step 5)
- `flagged_unpopulated_precincts.gpkg` (Step 6)

That flagging exists to surface precincts/blocks needing extra human scrutiny in the state-*provided* geometry itself (e.g. #283's `Monongalia_73` zero-population case). Once state-file corrections can change geometry -- not just names -- the flagging runs against a mutated file instead.

## Concrete example from this session

Monongalia_2A and Monongalia_2B don't exist in the as-provided state file at all. They're rows created by `R/result_analysis/WV_state_file_corrections/Monongalia_v1_reconciliation.r`, duplicating `Monongalia_2`'s geometry to match a county-level administrative relabeling (see #277). Every flagging/logging step downstream of Step 2 now evaluates population and block-overlap signals for two precincts that are a correction-script artifact, not something the state provided -- a case the original flagging logic (#285) wasn't designed around.

## Difficulty to be resolved

The above mentioned flagging/logging need to run against the as-provided file instead of the stable/corrected one, since their purpose is to surface issues in the state-provided data itself. 
Related: #277, #283, #285

Amasus:
Follow-up filed: #298 (heat maps on `feature/make_distance_heat_maps` need to account for the duplicate/dropped precinct geometry this ticket describes). Blocked by this ticket -- read this one's resolution first.

Amasus:
# Flagging Split and Step 7 Redesign Implementation Plan

[...]

Issue #277, opened 2026-06-21
[Precincts] Statewide WV precinct shapefile needs cleaning + version control (missing FIPS codes)


Amasus:
@Abd1tus — need your input on the versioning approach below.

## Finding

The statewide WV precinct shapefile (`datasets/precincts/.../VotingPrecincts_20260424_wmA84.shp`) has rows with no `StCoFIPS` value at all — confirmed during #266: Lewis County and Ohio County precincts have no FIPS code listed in the source data. This caused a real bug during Monongalia extraction (caught by review, not by luck): a naive `df[df$StCoFIPS == fips, ]` filter silently includes one all-`NA` placeholder row per `NA` in the comparison (since `NA == "54061"` is `NA`, not `FALSE`), inflating row counts in a way that's easy to miss without an explicit count assertion.

County name (`County_Nam`) data is consistently present across all rows, unlike `StCoFIPS`. #266 switched to filtering by county name instead of FIPS code for this reason — also consistent with this project's broader convention of keying most work off a `<County>_<ST>`-style location string (the Python config's `location` field) rather than FIPS codes.

## Implication: this file needs cleaning, which means we need version control

We'll need to clean the statewide file (at minimum: backfill or otherwise handle the missing Lewis/Ohio County FIPS codes) before it's fully reliable to use. That means we need an actual versioning strategy for it — right now `datasets/precincts/` is fully gitignored, so this file doesn't exist in version control at all.

**Proposed approach** (open for your input):
- Use git-lfs for the cleaned file. It's already a required project dependency (`CONTRIBUTING.md`) and already used for 316+ other files in this repo — no new tool needed. Note: shapefiles currently are NOT routed through LFS in `.gitattributes` (only marked `binary`, so line-ending conversion is skipped, but they're not LFS-tracked) — CSVs are the only file type currently going through `filter=lfs`. We'd want to fix that gap as part of this.
- Pair the LFS-tracked binary with a checked-in, version-controlled R script documenting exactly what cleaning was done (e.g. "filled missing `StCoFIPS` for Lewis/Ohio counties via `County_Nam` lookup"), so the transformation itself is reviewable via git diff/PR even though the shapefile binary isn't human-diffable.
- Considered and didn't recommend: GCS (this project's established use for it is generated outputs and DB credentials, not source-of-truth input data — using it here would separate data provenance from git history) and DVC (a new tool/dependency this project doesn't currently use, when git-lfs already solves the problem).

[...]

PR #300, opened 2026-07-11
feat(precincts): implement state-file versioning, resolve Monongalia v1 reconciliation (#277)

Amasus:
## Summary

Implements the state-file versioning scheme proposed on #277, and runs the process against the true as-provided WV statewide precinct file end-to-end (not the already-hand-patched file), testing the human-in-the-loop reconciliation workflow itself, not just the code.

- **Un-gitignored `datasets/precincts`, routed its shapefiles through git-lfs**, scoped to that path so `datasets/census/tiger`'s existing (non-LFS) shapefiles are unaffected.
- **Folder-per-step versioning**: `..._as_provided` (immutable original) → `..._Monongalia_County_WV_v1_reconciliation` → stable `West_Virginia_20260424_wmA84` (always mirrors the latest step; `STATE_PRECINCT_SOURCE_FILE` never changes). `next_correction_step_prefix()` derives the next step's name from what already exists on disk, so halt messages never point at a stale, already-applied correction script.
- **Generalized reconciliation to a single precinct-level gate.** `check_poll_precinct_agreement()` now merges state and county data by `Precinct_I` alone, so a pure rename produces one row with `state_USER_POLL_`/`USER_POLL_` side by side instead of two separate rows. This gate is a strict superset of what the old `match_location_names()` caught (verified empirically against the unfixed as-provided file), so that function is removed rather than left as a redundant second hard stop.
- **`Monongalia_v1_reconciliation.r`**: a one-time, dated correction record (not a reusable pipeline step) applying the human-reviewed corrections in `location_precinct_mismatches.csv` -- 6 pure renames via a plain join (no hardcoded values; the CSV is the reviewed record), the `Monongalia_2 → 2A/2B` county-level administrative relabeling (both new rows inherit `Monongalia_2`'s geometry, which is then dropped), and `Monongalia_A/B` dropped (non-populated, unassigned by the state, unreferenced by the county).
- **`assign_block_to_dominant_precinct()` now uses `with_ties = TRUE`**, so a block tied exactly between 2A/2B (identical duplicated geometry) is assigned to both rather than an arbitrary single winner via row order.
- Deleted two debug configs that pointed at folders removed earlier in this process, and the superseded `..._v1_fix_names` folder (replaced by `..._v1_reconciliation`).

## Verification

Ran `extract_precincts.r` against the as-provided file. It halted twice as expected -- once at the (then-separate) name gate, once at the precinct/ID gate -- each time reviewed with a human in the loop rather than reusing historical assumptions. After `Monongalia_v1_reconciliation.r`, a full end-to-end run completes with **zero reconciliation halts** (exit code 0), regenerating all four Step 4-7 output files. `Monongalia_73` still correctly flags as unpopulated -- the pre-existing, already-documented #283 case, unrelated to this work.

## Follow-ups filed, not addressed here

- #297: flagging/logging (this branch's own Step 4-7 outputs) now runs against the corrected state file rather than the as-provided original -- needs a design decision on whether/how to separate those.
- #298 (blocked by #297): the not-yet-merged `feature/make_distance_heat_maps` branch needs to account for duplicate (2A/2B) and dropped (A/B) precinct geometry before it merges.

Part of #277.


Amasus:
**Summary**

Replaces the hand-written, per-county precinct correction script with a generic, CSV-driven mechanism.

**What changed**
- `check_poll_precinct_agreement()`'s halt now writes two additional columns to `location_precinct_mismatches.csv` — `resolution_type` (rename/split/drop) and `geometry_source_precinct_I` — so every correction is captured as reviewed data instead of hardcoded per-county logic.
- New `build_precinct_crosswalk()` and `apply_corrections()` mechanically apply those reviewed decisions to the precinct shapefile (rename, split one precinct's geometry into multiple new IDs, or drop it entirely), and produce a `Precinct_I → resolved_polling_location` crosswalk artifact for anything downstream still holding a retired ID.
- `Monongalia_v1_reconciliation.r` — previously ~50 lines of county-specific hardcoded logic — now just calls `apply_corrections()`.

**A real gap the more thorough gate caught**

Re-running the gate against the untouched, as-provided state file surfaced a genuine mismatch (`Monongalia_2`, later split into `2A`/`2B`) that the original hand-maintained review record never captured. That's now correctly resolved as part of the reviewed CSV.

**Verification**

Rehearsed end-to-end from a genuine clean slate — reset to the as-provided state file, ran the full halt → human-review → correction pipeline fresh, and confirmed the corrected geometry and crosswalk output match the previously-shipped result exactly.

R/result_analysis/extract_precincts.r
antisocialscientist, line 169: 2 super-tiny comments that I should have made before:
1) Should I just not be picky about calling choropleth maps "heat maps" (probably a very minor concern, not worth the rewriting)
2) I have seen these before, but why are these 3 maps specifically hard coded in?

antisocialscientist (commented):
only one tiny question: Why are 3 specific maps hardoded at the bottom of extract_precincts.r?

As long as that is what you want, and I don't need to check the scratch_ files I am ready to approve the PR and handle the merges

R/result_analysis/extract_precincts.r
Amasus: extract precinct is an execution script. Those are the three demographic displays i want to show (for now) that list may change.

Issue #231, opened 2026-06-08
[driving-time] Phase 4: R analysis toolkit metric awareness


abd1tus:
Part of epic **epic-driving-time-metric**. Design context: #227. Depends on the metric data being available (Phases 1–3).

## Goal

Make the R analysis/visualization toolkit metric-aware so duration-metric runs render correctly (axis labels, units, summaries), not just distance.

## Changes

- Parameterize the metric column / units in the relevant `R/` analysis and plotting scripts (per `R/CLAUDE.md` conventions), so a run optimized on `duration_s` is labeled and summarized as time rather than distance.

## Tests

- `R/tests/r_smoke_test.R` passes; spot-check a duration-run plot/summary renders with correct units.

## Notes

- **Lowest priority / deferrable.** If scope needs trimming, this can ship after Phases 1–3 without blocking them. Default `driving_metric: distance` keeps all existing R output unchanged.

PR #313, opened 2026-07-14
feat(r): metric-aware graphs, maps, and regressions (#231)

Amasus:
## Summary
- Make `R/result_analysis/utility_functions/graph_functions.R` and `map_functions.R` metric-aware: a `driving_time`-optimized run now renders with time-based labels/units (minutes) instead of being mislabeled as distance in meters.
- `metric` is propagated as a per-row data column (via `select_varying_fields`/`create_descriptor_field`/`assign_descriptor_to_result`), with distance-bearing columns (`distance_m`/`weighted_dist`/`avg_dist`/`y_EDE`) scaled once at load time.
- Plot functions that combine descriptors on one shared axis warn-and-skip when the input mixes incompatible units (meters vs. minutes) rather than mislabeling or erroring; `haversine`/`driving_distance` (both meters) are treated as compatible.
- Map color scales (`distance_bounds()`) and the density-vs-distance regression plot (`plot_density_v_distance_bg`) compute bounds per unit instead of skipping, so mixed-metric config folders still render apples-to-apples comparisons within each metric.
- Dropped dead `log_flag`/`log_str` handling (a no-op since the code path that set it was commented out).

##Review note

Author has concerns about code quality. Feedback on how to refactor and improve data flow would be greatly appreciated.

## Test plan
- [x] Ran `R/result_analysis/Basic_analysis.r` end-to-end against `Monongalia_County_WV_driving_original_configs` (mixes `driving_distance` and `driving_time` configs in one folder) — completes with no errors.
- [x] Confirmed correct word/unit ("distance (m)" vs "time (min)") on generated plots and maps for each metric.
- [x] Confirmed mixed-metric plots warn and skip instead of erroring or mislabeling.
- [ ] `python run.py test` / `pylint python/` (no Python changes in this PR, but included per repo convention)

### DIFF
--- a/R/result_analysis/utility_functions/graph_functions.R
+++ b/R/result_analysis/utility_functions/graph_functions.R
@@ -690,19 +690,12 @@ plot_population_densities <- function(density_df){
 #at the block group level, ordered by population density
 #log / log scale, with best fit lines
 plot_density_v_distance_bg <- function(bg_density_data, county, demo_list, driving_flag = DRIVING_FLAG){
-	metric <- get_uniform_metric(bg_density_data, 'plot_density_v_distance_bg')
-	if (is.null(metric)) return(invisible(NULL))
-
-	#set graph y axis bounds. if min_distance == 0 m, make 1m
-	min_dist = min(bg_density_data[demographic %in% demo_list, ]$demo_avg_dist, na.rm = TRUE)
-	max_dist = max(bg_density_data[demographic %in% demo_list, ]$demo_avg_dist, na.rm = TRUE)
-	if (min_dist == 0){min_dist = min_dist + .01}
-    y_bounds = c(min_dist,max_dist)
-
 	#trim log density outliers
 	trimmed <- bg_density_data[abs(z_score_log_density)<4, ]
+	#tag each row with its metric's unit, so bounds can be computed per unit below
+	trimmed[, unit := sapply(metric, function(m) METRIC_LABELS[[m]]$unit)]
 
-    descriptor_graph <- function(descriptor_str, demo_list, y_bounds){   
+    descriptor_graph <- function(descriptor_str, demo_list, y_bounds, metric){
 		flag_strs <- make_flag_strs(driving_flag, metric)
 
 		title_str = paste0('Average', flag_strs$driving_str, flag_strs$word, ' to poll by demographic and block group')
@@ -719,7 +712,21 @@ plot_density_v_distance_bg <- function(bg_density_data, county, demo_list, drivi
 		add_graph_to_graph_file_manifest(graph_file_path)
 		ggsave(graph_file_path)
     }
-    descriptors = unique(trimmed$descriptor)
-    sapply(descriptors, function(x){descriptor_graph(x, demo_list, y_bounds)})
+
+	#set graph y axis bounds per unit (apples-to-apples within a unit;
+	#if min_distance == 0, make it .01 so log scale doesn't break)
+	plot_unit_group <- function(unit_data){
+		this_metric <- unit_data$metric[1]
+		demo_filtered <- unit_data[demographic %in% demo_list, ]
+		min_dist <- min(demo_filtered$demo_avg_dist, na.rm = TRUE)
+		max_dist <- max(demo_filtered$demo_avg_dist, na.rm = TRUE)
+		if (min_dist == 0){min_dist <- min_dist + .01}
+		y_bounds <- c(min_dist, max_dist)
+
+		descriptors <- unique(unit_data$descriptor)
+		sapply(descriptors, function(x){descriptor_graph(x, demo_list, y_bounds, this_metric)})
+	}
+
+	invisible(lapply(split(trimmed, trimmed$unit), plot_unit_group))
     }
 
--- a/R/result_analysis/utility_functions/graph_functions.R
+++ b/R/result_analysis/utility_functions/graph_functions.R
@@ -15,10 +15,11 @@ TABLES = c("edes", "precinct_distances", "residence_distances", "results")
 DEMO_COLS =  c("population", "hispanic","non_hispanic", "white", "black", "native", "asian", "pacific_islander", "other")
 
 PRINT_SQL = FALSE
-
-
-
-
+METRIC_LABELS <- list(
+	haversine        = list(word = 'distance', unit = 'm',   scale = 1),
+	driving_distance = list(word = 'distance', unit = 'm',   scale = 1),
+	driving_time     = list(word = 'time',     unit = 'min', scale = 1 / 60)
+)
 
 #########
 #Get flags from the config folders or config file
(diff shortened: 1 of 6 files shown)


Amasus:
Follow-up on the mixed-unit guard, addressing the code-quality concern raised above.

**Refactor: assembly-time checks instead of per-function guards**

Previously, each of 7 plotting functions independently called `get_uniform_metric()` and bailed with `invisible(NULL)` if the data mixed units. That's now computed once in `Basic_analysis.r`, right after the orig/potential data is read in:

- `resolve_flag_strs()` — checks a single folder's data for a uniform unit, returns the label strings or `NULL` (with the existing warning) if mixed.
- `resolve_combined_flag_strs()` — same, for the orig+potential combination.

`Basic_analysis.r` computes `orig_flag_strs`, `potential_flag_strs`, and `combined_flag_strs` once, and wraps each downstream plot call in `if (!is.null(...))`. The 7 plotting functions no longer guard themselves — they take `flag_strs` as a parameter and trust the caller. Maps and `plot_density_v_distance_bg` are intentionally not gated by this — they already handle mixed units correctly by splitting per-unit internally (per the earlier `1541ca19` fix), and don't consume these checks.

**Bug fix: silent mislabeling when two different metrics share a unit**

Found while re-reviewing this path: when a folder's data mixes two *different* metrics that happen to share the same unit (e.g. `haversine` + `driving_distance`, both meters), `resolve_flag_strs` was passing the full multi-value `metrics` vector into `make_flag_strs()`. `METRIC_LABELS[[metric]]` with a length-2 character vector triggers R's recursive `[[` indexing (`L[["a"]][["b"]]`) rather than erroring — so it silently returned `NULL`, and the resulting plot got a blank/malformed axis label (e.g. `"Equity weighted straight line  to poll"`, missing the word and unit) instead of a warning or a crash. Fixed by passing `metrics[1]` — safe here because `resolve_flag_strs` has already confirmed the group shares one unit, and every metric in `METRIC_LABELS` sharing a unit currently also shares the same `word`.

**Testing**

- Added a local-only test config (`Monongalia_County_original_driving_for_test.r`, not committed) that points `POTENTIAL_CONFIG_FOLDER` at the same mixed-unit folder as `ORIG_CONFIG_FOLDER`, to exercise the `potential_flag_strs`/`combined_flag_strs` paths that the existing historical-only Monongalia dataset doesn't reach.
- Ran `Basic_analysis.r` against it end-to-end: exactly the two expected warnings fire (`orig_data mixes units (driving_distance, driving_time)`, `potential_data mixes units (...)`), no spurious third warning for the combined check (correctly short-circuits when either side is already mixed), and none of the gated output files (`demographic_edes.png`, `avg_dist_distribution_boxplots.png`, `orig__y_EDE.png`, etc.) were produced. Maps, precinct persistence, and the per-unit-split density-vs-distance regression output still rendered correctly.
- Cross-checked every call site against its plotting function: every call gated behind a `flag_strs` check plots a distance/time-derived value (`y_EDE` or `avg_dist`); every ungated call outside maps/regression does not (`plot_precinct_persistence` plots `id_dest`/`demo_pop`, `plot_population_densities` plots `pop_density_km`) — confirming the gating boundary is drawn in the right place.

Amasus:
Update on branch state since the last comment (covers `6f61b0df`, `3f9cbd49`, `8ab37d5f` — all pushed).

**`METRIC_LABELS` as a `data.table` instead of a nested list**

`METRIC_LABELS` (metric → word/unit/scale) is now a `data.table` rather than a list of lists, matching this project's `data.table`-first convention. Every call site that did `sapply(metric, function(m) METRIC_LABELS[[m]]$field)` now does an update-join instead (`dt[METRIC_LABELS, unit := i.unit, on = 'metric']`), in `scale_distance_columns`, `resolve_flag_strs`, `plot_density_v_distance_bg`, and `distance_bounds` (map_functions.R).

`make_flag_strs` (the single-value lookup) changed shape along the way: it now filters the data.table (`metric_labels[metric %in% available_metrics, ]`) rather than doing `METRIC_LABELS[[metric]]`. Two real bugs surfaced and got fixed during this:
- A naive `metric_labels[metric == metric, ]` translation would have silently compared the table's own `metric` column to itself (always `TRUE`) rather than to the argument, because of a name collision — resolved by naming the parameter `available_metrics` instead of `metric`.
- `word` was being deduplicated (`unique(metric_info$word)`) before being returned, but `unit` wasn't — so a call matching more than one row (two different metrics sharing a unit, e.g. `driving_distance` + `haversine`) produced a length>1 `unit`, which `c(unit = ...)` silently splits into `unit1`/`unit2`, leaving `flag_strs$unit` `NULL` and the rendered label missing its unit entirely (e.g. `"...distance ()"`). Now both `word` and `unit` go through `unique()` symmetrically.

**Moved `scale_distance_columns` to where it's actually used across all four result tables**

It used to run once per table *inside* `assign_descriptor_to_result` (behind a `#TODO: I don't like this here`), a function whose job is otherwise just reading and merging in the descriptor/metric columns. It's now called once in `read_result_data`, right after the existing pass that merges in `num_polls`/`num_residences` — confirmed behavior-neutral (that merge is a pure row count, unrelated to the distance columns) and confirmed `assign_descriptor_to_result` has exactly one caller, so nothing else was relying on it returning pre-scaled data. Also added the doc comment `scale_distance_columns` was missing, and dropped a dead line in `load_results_from_csv` (`names(dt_list) <- names(file_path)` — redundant, since `lapply` over an already-named vector preserves the names).

**`upload_graph_files_to_cloud_storage()` re-enabled** in `Basic_analysis.r` — it had been temporarily commented out during local testing of the above.

**Testing**

Re-ran the full `Basic_analysis.r` end-to-end check against the mixed-unit Monongalia test config twice — once right after the `data.table` migration, once after the `scale_distance_columns` move — and confirmed identical results both times against the original baseline: the same two `mixes units` warnings, the same gated output files absent, the same per-unit-split density-vs-distance output present. Also ran `lintr::lint()` on both changed files (no new findings beyond a known `data.table` false-positive — `i.unit` flagged as "no visible binding," the same class of warning `lintr` already raises elsewhere in these files for other join-column references) and `R/tests/r_smoke_test.R` (passes).

R/result_analysis/utility_functions/graph_functions.R
```diff
@@ -340,14 +359,36 @@ demographic_legend_dict <- c(
...
+ 	return(as.list(c(driving_str = driving_str, word = unique(metric_info$word), unit = unique(metric_info$unit))))
+}
+
+#helper to guard plot functions against mixed-metric input data.
+#and assign correct flag strings accordingly.
+resolve_flag_strs <- function(df, data_str, driving_flag, metric_labels = METRIC_LABELS){
+	
+	#first, get check if the data mixes units. 
+	metrics <- unique(df$metric)
+	units <- unique(metric_labels[.(metrics), on = 'metric', unit])
```
antisocialscientist, line 380: I have been trying to figure out whether I don't quite understand here, but I worry about whether it is supposed to accept multiple metrics which share units

antisocialscientist: actually, i think I am just being a bozo here

antisocialscientist:
I really don't understand why this isn't a good PR. I spent a lot of time looking at it and am not seeing the fear from the review flag anywhere. I am assuming that it was addressed by subsequent commits?

Issue #226, opened 2026-06-01
Mitigate silent cross-border distance inflation from single-state ORS extracts


abd1tus:
## Context

The driving-distance toolset builds a routing graph from a **single US state** Geofabrik extract (`<state>-latest.osm.pbf`) and computes census-block-centroid → polling-place distance matrices against it. Research (`docs/ors_cross_state_research.md`, Findings B/D) surfaced a silent accuracy problem at **state borders**.

Geofabrik state extracts use a **complete-ways** clip (border-crossing ways are kept whole) but **no buffer** — cross-streets, interchanges, and return loops on the far side of the line are dropped. So when the genuinely shortest drive between two in-state points would briefly cross into a neighboring state (an interstate that clips the line, a river bridge just over the border, a jagged boundary), those out-of-state road segments are not in the graph.

**The failure mode is silent.** Unlike a true no-route cell (which ORS returns as JSON `null` — confirmed empirically for v8.0.0, see Finding A), a border-forced detour comes back as a **valid, normal distance in meters, HTTP 200, no flag**. ORS faithfully returns the best path *on the graph it was given* — a longer route that winds down in-state roads to stay inside the boundary polygon. The number is real; it just doesn't reflect how anyone would actually drive.

**Why this matters more here than for a generic router.** This is a Kolm-Pollak *equity* optimizer. A border detour does not produce random noise — it produces **spatially clustered, systematic bias**: every census block near that stretch of border is inflated in the same direction. The model then concludes those communities have worse access than they do, skewing which areas the optimizer prioritizes. A coherent geographic bias is the most dangerous kind for an equity tool, and it is invisible in aggregate statistics.

**Scope / severity.** Impact is entirely a function of *which county is analyzed*:
- **Interior counties** (e.g. Gwinnett_GA, well inside the state): effectively zero impact.
- **Border-adjacent counties** (e.g. a county on the Savannah River across from SC, or on the GA/AL line): potentially material, concentrated in the border-proximate blocks.

This ticket is for **deciding and implementing a mitigation** (or formally documenting the limitation). It is distinct from #225 (which guards against negative/invalid distance values).

## A note on where the cost actually lives

[...]

PR #321, opened 2026-07-20
feat(ors): driving-distance tooling + cross-border buffered extract (#226)

### LINKED ISSUE #216: Driving-distance tools: integrate upstream geolib functionality
## Goal

Absorb the upstream [`dxwils3/geolib`](https://github.com/dxwils3/geolib) driving-distance functionality into this project as a first-class toolset. Replaces the previous out-of-repo dependency with a config-driven CLI plus the sibling Docker container that runs OpenRouteService (ORS) locally.

End state: `python run.py generate_driving_distances_cli -l <config>` reads census-block centroids (origins) and the potential-polling-locations CSV (destinations) for a county, routes every pair through a local ORS instance, and writes the canonical 3-column long-form CSV that the solver already consumes.

## Branch & base

- Branch: `feature/driving-distance-tools`
- Base: `feature/auth-credentials` (transitive: end-to-end pipeline `pull_census_data` → `generate_driving_distances_cli` → solver needs the census key infrastructure that branch establishes)

## What got built

**Three utility modules** (`python/utils/`):
- `ors_url.py` — URL resolution (CLI override > `\$ORS_URL` > default) and matrix↔directions URL derivation.
- `ors_client.py` — thin HTTP wrapper around ORS matrix and directions endpoints. Raises `OrsMatrixError` on API errors; `query_directions` returns `None` on malformed payloads.
- `driving_distance_matrix.py` — orchestration: batching (≤ 2,500 ORS cells / call, ≤ 10 sources/batch), per-source retry on `OrsMatrixError`, haversine snap (< 1 km) for unroutable origins, partial-CSV resume.

**Four CLI scripts** (`python/scripts/`):
- `generate_driving_distances_cli` — main CLI; reuses `model_data.py` helpers to derive origins/destinations from a county config; tees stdout + log file.
- `ors_setup_cli --state <CODE>` — host-only Geofabrik downloader to `.devcontainer/ors_data/`.
- `ors_up_cli` — host-only `docker compose up -d` with health-poll readiness wait (45-min timeout, dumps last 50 container log lines on timeout).
- `ors_down_cli [--purge-graphs]` — host-only stop, optionally removes the named volume.

**Sibling docker-compose** (`.devcontainer/docker-compose.ors.yml`):
- Pinned `openrouteservice/openrouteservice:v8.0.0`.
- Port mapping `8080:8082` (ORS v8 listens on 8082 internally).
- Bind-mount `.devcontainer/ors_data/` → `/home/ors/files/`.
- Named volumes for built graph, elevation cache, logs.
- Explicit `ors.engine.source_file` env var (ORS auto-detect picks up its own demo `.osm.gz` if not constrained).
- XMS/XMX env vars (image's launcher ignores JAVA_OPTS for heap settings).
- Docker healthcheck mirroring `ors_up_cli`'s poll.

**Glue**:
- `.devcontainer/.env` with `COMPOSE_PROJECT_NAME=polling` so app and ors share a Docker network.
- `.devcontainer/docker-compose.yml` updated: `working_dir: /workspaces/Equitable-Polling-Locations`, `ORS_URL=http://ors:8082/...` so app container reaches ors by service name.
- `run.py` special-cases ORS lifecycle commands (run on host as file paths, bypassing the docker-compose wrapper).
- `.gitignore` updated for `.devcontainer/ors_data/*.pbf` plus explicit `!.devcontainer/.env` negation.
- `docs/to_run.md` gained a 'Generating driving distances' workflow section.

**Tests**: six new files (`python/tests/`), 37 unit tests, stdlib `unittest.mock` only. Full project suite goes from 218 → 220 passing (the +2 are the resume helper's tests). 10/10 pylint on every new file.

## Key technical decisions

- **Layered design**: CLI → orchestration (`driving_distance_matrix`) → service wrapper (`ors_client`). Each layer mocks only the boundary directly below it in tests. If we ever swap ORS for Valhalla/OSRM/Mapbox, only `ors_client` moves.
- **Config-driven CLI input** (not a `--sources-file` / `--destinations-file` pair): reuses the solver's own data layer, single source of truth for origin/destination sets.
- **Local-only ORS** in v1: no cloud-endpoint config, no API key plumbing, no `ors_key` in credentials.json. Filed as deferred.
- **Stdlib-only for host-side scripts**: lifecycle commands (`ors_setup_cli`, `ors_up_cli`, `ors_down_cli`) use `urllib.request` and `subprocess` only — no `requests` import — so they run with the host's plain `python3` regardless of pip state. `run.py` invokes them as file paths so the `python/__init__.py` package cascade (which pulls in `numpy` via the database module) doesn't trigger.
- **Output CSV strips provenance column** that geolib's upstream code emits, matching the existing project consumer's 3-column shape (`id_orig,id_dest,distance_m`).
- **Partial-CSV resume**: the upstream geolib bug where existing partial output caused the CLI to skip work was fixed; on re-run, the CLI subsets to the not-yet-fetched pairs and merges with the existing rows.
- **`example-heidelberg.osm.gz`** workaround: ORS v8 writes a demo `.osm.gz` into the bind-mounted directory on first boot, and its auto-detect prefers that over the `.pbf` we downloaded. The compose file's `ors.engine.source_file` env var forces our file.

## Smoke-tested end to end

Two real-world smoke tests on host hardware:

1. **`testing` location** (~25 census blocks × ~15 destinations, Gwinnett GA subset). ✓ Wrote `datasets/driving/testing/testing_driving_distances.csv` with real Atlanta-area road distances.
2. **`Tarrant_County_TX_year_2024`** (31,925 census blocks × 368 destinations = 11.7M cells). ✓ 7-minute matrix run produced a 603 MB output CSV at `datasets/driving/Tarrant_County_TX/Tarrant_County_TX_driving_distances.csv` with real DFW road distances (AT&T Stadium etc.).

Multi-state was exercised by switching Georgia → Texas: `ors_down_cli --purge-graphs`, edit `ors.engine.source_file` filename in the compose, `ors_setup_cli --state TX`, `ors_up_cli`. Texas graph build took ~15 min with `XMX=10g`.

## Documentation strategy

Per `CLAUDE.md` policy, the design spec (`docs/superpowers/specs/2026-05-21-driving-distance-tools-design.md`) and implementation plan (`docs/superpowers/plans/2026-05-21-driving-distance-tools.md`) are kept as local working files (`docs/superpowers/` is gitignored). The substance from both is in this issue and in the commit messages on the branch.

## Follow-ups filed as separate issues under this milestone

- Per-batch progress logging in `build_distance_matrix` orchestration (smoke test exposed a 7-minute silent wait).
- Auto-detect the `.pbf` in `.devcontainer/ors_data/` instead of hardcoding the filename (current multi-state UX requires editing the compose file).
- Wire `--verbose` / `-vv` flag on `generate_driving_distances_cli` to actually gate screen output (currently parsed but unused).

## Status

Implementation complete. 26 commits on `feature/driving-distance-tools`. Branch is local-only (unpushed) pending review/merge decision.

### LINKED ISSUE #217: Add per-batch progress logging to driving_distance_matrix orchestration
## Context

During the Tarrant_County_TX smoke test (#216), `generate_driving_distances_cli` ran 6,400+ ORS matrix batches over ~7 minutes and printed **no progress** during that time. The user saw `resume: 0 rows present, 11748400 pairs to fetch` and then nothing until the final `wrote 11748400 rows to ...` line. The log file had only 6 lines total (start, paths, summary).

This is a UX gap, not a correctness gap — the matrix calls all ran successfully.

## Root cause

When the upstream `geolib/driving_distances.py::get_all_distances` was refactored into the layered `python/utils/driving_distance_matrix.py::build_distance_matrix`, the per-batch progress print:

```python
print(f'{starting_index}/{len(source_names)}: {duration:.4f} seconds')
```

was dropped. The orchestration layer no longer emits anything between batches.

## Fix

Add per-batch progress reporting in `build_distance_matrix` (or in `_fetch_one_batch` if cleaner). Options:

1. Accept an optional `progress_callback(batch_index, total_batches, elapsed_s)` kwarg; the CLI passes a callback that tees to stdout + log file via its existing `_tee` helper.
2. Use `tqdm` directly (already a project dependency via geolib's lineage and now pinned in `environment.yml`).

Option 1 keeps the orchestration layer free of UX coupling and is more testable; option 2 is faster to write. Either way, also emit:

- Snap events (`unroutable origin <id>: snapped to <neighbor> (haversine = N m)`) — currently silent.
- Per-source retry messages (`retrying source <id> individually`) — currently silent.

## Acceptance

- A 1+ minute run of `generate_driving_distances_cli` against any non-trivial config prints periodic progress to both stdout and the log file.
- Existing tests still pass (the change should not affect public API of `build_distance_matrix`).
- New test verifies the callback is invoked the expected number of times (or that tqdm output appears in captured stderr if going that route).

### LINKED ISSUE #218: Auto-detect .pbf in .devcontainer/ors_data/ instead of hardcoding state filename
## Context

Today, `.devcontainer/docker-compose.ors.yml` has the .pbf path hardcoded:

\`\`\`yaml
environment:
  ors.engine.source_file: /home/ors/files/georgia-latest.osm.pbf
\`\`\`

Switching states requires editing this YAML, which forces a workflow like:

\`\`\`bash
python3 run.py ors_down_cli --purge-graphs
python3 run.py ors_setup_cli --state TX
# Edit docker-compose.ors.yml -> change filename to texas-latest.osm.pbf
python3 run.py ors_up_cli
\`\`\`

This was an explicit v1 trade-off recorded in #216 (favor a working smoke test over a flexible state-switch UX), but it bites every multi-state user.

## Why the obvious fix doesn't work

ORS v8's auto-detection scans `/home/ors/files/` for `.osm.*` files, but the image **writes its own `example-heidelberg.osm.gz`** into that directory on first boot. Auto-detection then prefers the demo file (or both files at once, indeterministically), which is why the explicit `source_file` env var was added in the first place.

## Proposed fix

Have `ors_up_cli` scan `.devcontainer/ors_data/` for `.osm.pbf` files at lifecycle time and inject the path via env-var substitution in compose:

`.devcontainer/docker-compose.ors.yml`:
\`\`\`yaml
environment:
  ors.engine.source_file: /home/ors/files/\${ORS_PBF_FILENAME:?ors_up_cli must set ORS_PBF_FILENAME}
\`\`\`

`python/scripts/ors_up_cli.py`:
\`\`\`python
def _find_pbf(ors_data_dir: Path) -> str:
    pbfs = sorted(ors_data_dir.glob('*.osm.pbf'))
    if not pbfs:
        raise FileNotFoundError(
            f'No .osm.pbf in {ors_data_dir}. Run \\\`python run.py ors_setup_cli --state X\\\` first.'
        )
    if len(pbfs) > 1:
        print(f'Warning: multiple .pbf files found; using {pbfs[0].name}')
    return pbfs[0].name

# In main():
env = os.environ.copy()
env['ORS_PBF_FILENAME'] = _find_pbf(...)
subprocess.run(['docker', 'compose', ..., 'up', '-d'], env=env, check=True)
\`\`\`

End state: user runs `ors_setup_cli --state TX`, then `ors_up_cli`. No compose edit. Switching states is just rerunning `ors_setup_cli` (with `--force` if needed) plus `ors_down_cli --purge-graphs` to drop the cached graph.

## Acceptance

- Switching from Georgia to Texas requires zero file edits.
- `ors_up_cli` errors clearly if `.devcontainer/ors_data/` is empty.
- Tests cover the lookup helper and the env-var passing.
- `docker-compose.ors.yml` no longer references any specific state filename.

### LINKED ISSUE #219: Wire -v / --verbose flag in generate_driving_distances_cli to actually gate output
## Context

`generate_driving_distances_cli` accepts `-v` / `--verbose` (count, default 0), parsed via argparse, but **nothing in the code reads `args.verbose`**. Passing `-v` or `-vv` is currently a no-op.

The intent (recorded in the spec) was: the log file always captures everything; the `-v` count gates how chatty the screen output is — e.g. `-v` prints batches, `-vv` adds per-cell info.

## Fix

Either:

1. Pass `args.verbose` into `_tee()` as a threshold; each `_tee` callsite tags itself with a 'verbosity level' (the screen gets it only if `verbosity_level <= args.verbose`).
2. Switch to Python's `logging` module with WARNING/INFO/DEBUG levels; `args.verbose` maps to handler level.

Option 1 keeps the simple `print()`-and-tee model the rest of the project uses. Option 2 is more idiomatic Python but introduces a logging pattern not used elsewhere in this codebase.

Probably option 1, with two screen-relevant levels:
- Default (0): start/finish lines, errors, snap events only.
- `-v` (1): + per-batch progress (see #217).
- `-vv` (2): + per-source retry, per-cell info if any.

## Acceptance

- Running `generate_driving_distances_cli ... -vv` produces strictly more screen output than `... -v` than no flag.
- Log file is unchanged across all verbosity levels (still captures everything).
- A test verifies the gating by patching `_tee` and asserting calls with `to_screen=False` get filtered when verbosity is low.

## Related

Naturally pairs with #217 (per-batch progress logging) — the progress lines should be `-v`-gated to screen.

### LINKED ISSUE #220: Cleanup: replace remaining bare column-name strings in driving-distance code with DISTANCE_* constants
## Context

The verbosity work in #217 / #219 replaced bare `'id_orig'` / `'id_dest'` / `'distance_m'` literals with the canonical `DISTANCE_*` constants from `python/solver/constants.py` in every function it touched (`build_distance_matrix`, `_retry_sources_individually`, `_snap_unroutable_origins`, the CLI's `OUTPUT_COLUMNS`). The boy-scout rule stopped at the function boundary — untouched helpers in the same modules still use bare strings.

This issue captures the remaining cleanup so future readers don't have to wonder which strings are "data we should constant-ize" vs which are "okay as literals."

### `python/utils/driving_distance_matrix.py`

| Line | Function | Code |
|---|---|---|
| 67 | `get_missing_origins` | `df[pd.isnull(df['distance_m'])]` |
| 68 | `get_missing_origins` | `set(missing_rows['id_orig'])` |
| 86 | `matrix_response_to_long_df` | `rows.append({'id_orig': source, 'id_dest': dest, 'distance_m': distance_m})` |
| 87 | `matrix_response_to_long_df` | `pd.DataFrame(rows, columns=['id_orig', 'id_dest', 'distance_m'])` |
| 109-128 | `estimate_origin` | Multiple bare references across rename/filter/group-by/return slices |
| 276 | `build_distance_matrix` (snap branch) | `set(df.loc[~pd.isnull(df['distance_m']), 'id_orig'])` |
| 281 | `build_distance_matrix` (snap branch) | `df = df.dropna(subset=['distance_m'])` |
| 349 | `build_distance_matrix` (empty-fallback) | `pd.DataFrame(columns=['id_orig', 'id_dest', 'distance_m'])` |
| 386 | `resume_from_partial_output` | `pd.DataFrame(columns=['id_orig', 'id_dest', 'distance_m'])` |
| 388 | `resume_from_partial_output` | `set(zip(existing_df.get('id_orig', []), existing_df.get('id_dest', [])))` |

### `python/scripts/generate_driving_distances_cli.py`

| Line | Function | Code |
|---|---|---|
| 211 | `main` (--check-bad-locations branch) | `set(source_ids) - set(df['id_orig'])` |
| 244 | `main` (resume path) | `combined.drop_duplicates(subset=['id_orig', 'id_dest'], keep='last')` |

### Other repo-wide candidates (NOT in scope for this ticket)

A repo-wide `grep "'id_orig'\\|'id_dest'\\|'distance_m'"` would also turn up hits in `python/solver/model_data.py`, `python/database/models/driving_distances.py`, and various places that build or consume distance-data frames. Those are *out of scope* for this ticket — the goal here is to finish the bare-string cleanup specifically in the two files the verbosity work touched, plus their tight neighbors in `driving_distance_matrix.py`. A separate ticket can sweep the rest of the repo if desired.

Test files (`python/tests/driving_distance_matrix_test.py`, `python/tests/generate_driving_distances_cli_test.py`) typically construct fixture data with literal strings (`{'id_orig': [...]}`). Per project memory `feedback_constants_over_string_literals.md`, those CAN stay literal — they're test inputs, not shared concepts. Switching them is optional taste, not a requirement of this ticket.

## Also: `_OUTPUT_COLUMNS` duplication

`python/utils/driving_distance_matrix.py` defines `_OUTPUT_COLUMNS = [DISTANCE_ID_ORIG, DISTANCE_ID_DEST, DISTANCE_DISTANCE_M]` (private). The CLI defines an identical public `OUTPUT_COLUMNS` separately. Two reasonable resolutions:

- **A.** Leave both (matches the "one bad convention beats two" principle — each module is self-contained). Document the duplication intentionally.
- **B.** Make the library's `_OUTPUT_COLUMNS` public (drop the underscore) and have the CLI import it. Fewer constants but cross-module coupling.

Pick one as part of this cleanup.

## Acceptance

- Every bare `'id_orig'` / `'id_dest'` / `'distance_m'` reference in `python/utils/driving_distance_matrix.py` and `python/scripts/generate_driving_distances_cli.py` is replaced with the corresponding `DISTANCE_*` constant.
- `_OUTPUT_COLUMNS` / `OUTPUT_COLUMNS` duplication is resolved one way or the other.
- All existing tests still pass.
- Pylint 10/10 on touched files.
- Single commit titled `refactor(utils,scripts): replace remaining bare column strings with DISTANCE_* constants`.

## Related

- Depends on #216 / #217 / #219 (already closed). Final follow-up under the epic.

### LINKED ISSUE #223: Replace runtime /ors/v2/status verification with per-state graph cache isolation
## Context

In #218 we added `verify_loaded_state` (`python/utils/ors_status.py`) to protect against the failure mode where the cached routing graph in the `ors_graphs` named volume doesn't match the requested state.

Smoke testing on the host (against ORS v8.0.0) revealed that the assumption baked into this check — that `/ors/v2/status` exposes a `source_file` field — is wrong. The v8.0.0 status response has no field identifying which OSM extract is loaded. Sample response (with `creation_date: ""` and no source_file / extent / bbox anywhere):

```json
{
    "engine": {"build_date": "2024-03-21T14:04:52Z", "version": "8.0.0"},
    "profiles": {
        "profile 1": {
            "storages": { /* gh_profile entries */ },
            "profiles": "driving-car",
            "creation_date": "",
            "limits": { /* numeric caps */ }
        }
    },
    "services": ["routing", "isochrones", "matrix", "snap"]
}
```

Research against the upstream project (https://github.com/GIScience/openrouteservice) confirmed:

1. ORS v9.x doesn't add introspection either; the upstream stance is that the routing engine is intentionally unaware of its data origins.
2. Graph cache reuse behavior is unchanged across 8.x→9.x: if a built graph is present in `/home/ors/graphs`, ORS uses it regardless of `source_file`.

The structural fix is to **bind-mount the graphs directory to a state-specific host path**. Each state gets its own cache dir; swapping states automatically maps to a different host dir → forces a rebuild. No runtime verification needed.

## Changes

- Replace the named `ors_graphs:` volume in `.devcontainer/docker-compose.ors.yml` with `../datasets/ors_graphs/${ORS_STATE:?...}:/home/ors/graphs`
- Rename the substitution variable from `ORS_PBF_FILENAME` to `ORS_STATE` (compose now derives the .pbf filename via `${ORS_STATE}-latest.osm.pbf`)
- Delete `python/utils/ors_status.py` and `python/tests/ors_status_test.py`
- Remove the verify call, import, and importlib fallback from `python/scripts/ors_up_cli.py`
- Remove the verify-related tests from `python/tests/ors_up_cli_test.py`
- Drop `--purge-graphs` from `ors_down_cli` (the named volume escape hatch no longer applies; document `rm -rf datasets/ors_graphs/<slug>` instead)
- Update `docs/to_run.md` — "Switching states" no longer needs `--purge-graphs`
- Fix `run.py` orchestration teardown bug: wrap spawn+matrix in one try/finally so an `ors_up_cli` failure still triggers `ors_down_cli`

## Out of scope

- ORS v9.x upgrade (separate follow-up; involves config namespace migration `ors.engine.source_file` → `ors.engine.profile_default.build.source_file`, JVM heap revalidation, and graph cache behavior re-test)

## Migration note

Anyone with an existing local `polling_ors_graphs` named volume can `docker volume rm polling_ors_graphs` after this lands; it's no longer referenced. The new per-state cache directories are auto-created on first use under `datasets/ors_graphs/<slug>/`.

### LINKED ISSUE #224: Derive ORS state from config location instead of requiring a CLI arg
## Context

`generate_driving_distances_cli` currently requires a positional `<state>`
slug:

```
python3 run.py generate_driving_distances_cli georgia -l <config>
```

Every config that uses this CLI already names a location (e.g.
`Tarrant_County_TX`); the project's naming convention is `<Name>_<ST>`
where `<ST>` is the 2-letter postal code. Requiring the user to also type
the state is redundant — derive it from `config.location`.

## Changes

- `--state <slug>` becomes an optional override (the typical invocation
  drops it entirely)
- New helpers in `python/utils/ors_setup.py`:
  `STATE_CODE_TO_SLUG`, `state_slug_from_location`,
  `location_from_config_file`
- `generate_driving_distances_cli` and `run.py` orchestration both
  derive state from the config's location when `--state` is absent
- `run.py`'s `_peel_orchestration_args` helper is deleted; its job
  becomes four lines of inline `parse_known_args`
- For synthetic configs whose location doesn't end in `_<ST>`
  (`testing`, `e2e_*`), users pass `--state` explicitly

Spec: see `docs/superpowers/specs/2026-05-31-derive-state-from-config-design.md`
(locally gitignored).

### LINKED ISSUE #225: Guard against negative (< 0) driving distances at solver ingestion and ORS generation
## Context

While reviewing cross-state ORS routing (see `docs/ors_cross_state_research.md`) we confirmed empirically that ORS **v8.0.0** `/v2/matrix/driving-car` with `metrics:["distance"]` returns JSON **`null`** (not `-1`) for an unroutable cell — so there is no live source of negative distances today. The original `-1` concern (raised by web research) was refuted against the live georgia graph.

However, the review also found that **no layer rejects a negative distance** if one ever appeared:

- `query_matrix` (`python/utils/ors_client.py`) only checks that a `distances` field exists.
- `matrix_response_to_long_df` (`python/utils/driving_distance_matrix.py`) writes each value into `distance_m` verbatim.
- `get_missing_origins` / `_snap_unroutable_origins` key off `pd.isnull` only.
- The solver ingestion gate at `python/solver/model_data.py:750` (`filter_distance_data`) also checks only `pd.isnull`.

**Impact if a negative ever appears** (a future ORS image bump, an alternate metric, the DB import, or a legacy/hand-edited CSV): it passes every check, lands in `distance_m`, and is written/consumed as a valid distance. Because the Kolm-Pollak objective **minimizes** distance, a `-1 m` "drive" is *more* attractive than a real `0 m` one — the optimizer would actively prefer that origin→destination pairing. Silent and high-impact.

We are protected today only by the empirical fact that v8.0.0 emits `null`; there is no defense-in-depth behind it. This ticket adds two complementary guards.

## Changes

**Guard 1 — solver ingestion (the backstop), `model_data.py` `filter_distance_data` (~line 750).** Protects every data source (this CLI, DB import, legacy/hand-edited CSVs). Widen the existing missing-distance gate so that `invalid = pd.isnull(distance_m) | (distance_m < 0)`, compute the offending origin/destination sets from the *valid* subset (`notna & >= 0`), and raise the same-style `ValueError` with the message widened to "missing or invalid (negative)". `0` stays valid (legitimate same-point distance). One code path, consistent with the existing hard-fail-on-bad-data convention.

**Guard 2 — generation boundary, `python/utils/driving_distance_matrix.py`.** Protects the persisted CSV. In `build_distance_matrix`, after the matrix + directions-retry DataFrame is assembled and **before** `_snap_unroutable_origins`, coerce `distance_m < 0` → `NaN`. Negatives then flow through the existing `get_missing_origins` → snap/retry machinery exactly like a real no-route `null`, so nothing negative is ever written to the output CSV. (Placing the coercion before snapping is what lets the graceful recovery path handle it.)

**TDD (per `python/CLAUDE.md`): failing tests first.**
- Solver: a filtered distance DataFrame containing a `-1` raises `ValueError` and names the offending origin/destination.
- Generation: a matrix response containing a negative cell is treated as unroutable (snapped or dropped), never emitted as a negative value.

## Coverage — what these guards do and do NOT catch

**Guard 1 validates only the subset of rows the current config uses.** Rows dropped *before* the line-750 check escape it:

- **`config.bad_types` exclusions** (`model_data.py:737`) — e.g. config sets `bad_types: [EV_2020]` and a block→`EV_2020` distance is `-1`; that row is dropped before the check.
- **Alpha-calibration pass** (`for_alpha=True`, lines 711–718) strips **all** potential locations and centroids — e.g. block→`potential_school_5` `= -1` is invisible to the alpha pass (the base pass would still catch it).
- **Year / dest_type filtering** (`filter_dest_type`, line 739) — e.g. CSV holds 2016 + 2020, config runs `year: [2020]`, and a `-1` on a 2016 row is never checked this run.

In every escape case the negative is on a row the run **doesn't use**, so it cannot corrupt that run's optimization — Guard 1 fires precisely when a negative could affect the result. The residual risk is that a negative can sit **undetected in the CSV** until some *future* config selects that row. **Guard 2 closes exactly that gap** by keeping negatives out of the file in the first place. The two guards are complementary: **Guard 1 protects the run, Guard 2 protects the file.**

## Out of scope

- Broadening Guard 1 to scan the entire *unfiltered* distance file. Kept consistent with the existing per-config null check; Guard 2 covers file-level cleanliness instead.
- The border-distance **accuracy** issue (valid-but-inflated in-state detours when a single-state extract lacks a cross-border segment) — separate concern, tracked in `docs/ors_cross_state_research.md` Findings B/D.

## Reference

- `docs/ors_cross_state_research.md` — Finding A (empirical `null`-not-`-1` result) and the negative-distance guard design + coverage analysis.

### LINKED ISSUE #305: generate_driving_distances_cli: resume_from_partial_output never resumes, corrupts output with duplicate rows
## Summary

`resume_from_partial_output` treats every `(id_orig, id_dest)` pair as missing on every run, because of an `int` vs. `str` dtype mismatch — so "resume" always does a full re-fetch, and the subsequent dedup step fails to collapse the resulting duplicates. The output CSV ends up with **every pair duplicated**: one stale row (old schema, `duration_s` empty) and one fresh row (new schema, `duration_s` populated).

## Root cause

`resume_from_partial_output` ([python/utils/driving_distance_matrix.py:412-443](../blob/main/python/utils/driving_distance_matrix.py#L412)):

```python
existing_df = pd.read_csv(output_path)
...
present = set(zip(
    existing_df.get(DISTANCE_ID_ORIG, []),
    existing_df.get(DISTANCE_ID_DEST, []),
))
all_pairs = [(s, d) for s in source_ids for d in dest_ids]
remaining = [pair for pair in all_pairs if pair not in present]
```

- `pd.read_csv` infers `id_orig` as `int64`, since census block GEOIDs look numeric.
- `derive_origins_and_destinations` in [python/scripts/generate_driving_distances_cli.py:103](../blob/main/python/scripts/generate_driving_distances_cli.py#L103) builds `source_ids` as `str(row[TIGER20_GEOID20])`.
- So `present` holds `(540610101011000, "some location")` (int, str) tuples, while `all_pairs` holds `("540610101011000", "some location")` (str, str) tuples. These never compare equal, so **every pair looks "remaining"** even when the CSV already has 100% coverage — the resume optimization never resumes.

The same dtype mismatch then breaks dedup at [python/scripts/generate_driving_distances_cli.py:320-323](../blob/main/python/scripts/generate_driving_distances_cli.py#L320):

```python
combined = pd.concat([existing_df, new_df], ignore_index=True)
combined = combined.drop_duplicates(
    subset=[DISTANCE_ID_ORIG, DISTANCE_ID_DEST], keep='last',
)
```

`pd.concat` mixes the `int64` values from `existing_df` with `str` values from `new_df` into an `object` column, so `540610101011000` (int) and `"540610101011000"` (str) are treated as distinct keys and never deduplicated.

## Repro

Observed on `feature/driving-time-metric`, running `generate_driving_distances_cli` against the already-populated `datasets/driving/Monongalia_County_WV/Monongalia_County_WV_driving_distances.csv` (a legacy 3-column file: `id_orig,id_dest,distance_m`, no `duration_s`):

- Input: 66,528 data rows (3-column, no `duration_s`).
- Output after running the CLI: 133,056 data rows — exactly double.
- Every `(id_orig, id_dest)` pair appears twice: once with `duration_s` empty (carried over verbatim from the old file) and once with `duration_s` populated (freshly fetched).

```
$ wc -l Monongalia_County_WV_driving_distances.csv
133057  # 133056 rows + header, was 66529 (66528 rows + header)

$ awk -F, 'NR>1{print $1","$2}' Monongalia_County_WV_driving_distances.csv | sort | uniq -c | sort -rn | head -3
      2 540610120004019,University High School
      2 540610120004019,Triune-Halleck VFD
      2 540610120004019,Town of Granville Social Hall
```

## Impact

- The "resume from partial output" feature (added specifically to warm-start from an existing file) is currently a no-op — it always does a full re-fetch.
- Any state whose block GEOIDs have a leading zero (e.g. Alabama, `01...`) would silently lose that leading zero when `pd.read_csv` infers `id_orig` as `int64`, corrupting the ID rather than just duplicating rows.
- Downstream, a driving-distances CSV with duplicate `(id_orig, id_dest)` pairs is invalid input for the solver, which expects one row per pair.

## Suggested fix

Force `id_orig` (and ideally `id_dest`, for symmetry) to `str` when reading the CSV back in `resume_from_partial_output`, e.g. `pd.read_csv(output_path, dtype={DISTANCE_ID_ORIG: str, DISTANCE_ID_DEST: str})`, so the dtypes match `source_ids`/`dest_ids` on both the "present" check and the later `drop_duplicates` call. Add a regression test that seeds a partial CSV with numeric-looking IDs and asserts `resume_from_partial_output` recognizes them as already present.

## Workaround used

Deleted the stale `Monongalia_County_WV_driving_distances.csv` and re-ran `generate_driving_distances_cli` from scratch (no partial file to resume from, so the bug doesn't trigger on a clean run).

### LINKED ISSUE #320: ORS container creates root-owned datasets/openrouteservice/, breaking git operations for anyone who's run it
## Summary

`docker-compose.ors.yml`'s `ors` service (`openrouteservice/openrouteservice:v8.0.0`) runs as root with no `user:` override, and bind-mounts two host directories:

```yaml
volumes:
  - ../datasets/openrouteservice:/home/ors/files
  - ../datasets/ors_graphs/${ORS_STATE}-buffered:/home/ors/graphs
```

If `datasets/openrouteservice/` doesn't already exist on the host when `docker compose up` first runs, the Docker daemon auto-creates the bind-mount point as **root-owned**. The container then writes into it as root too — confirmed directly: the image writes `example-heidelberg.osm.gz` into `/home/ors/files` on first boot (see the existing comment in `docker-compose.ors.yml` about needing `ors.engine.source_file` set explicitly to avoid this), and that file lands root-owned on disk.

Once `datasets/openrouteservice/` is root-owned, the local dev user can't create or delete anything inside it (Unix requires write on the *parent* directory to add/remove entries — the directory's own `drwxr-xr-x root:<group>` has no group-write bit). This breaks:
- `git checkout` / `git pull` / `git merge` whenever a branch tracks a file at that path (e.g. the `.gitkeep` scaffolding added in #319) — fails with `unable to unlink`/`unable to create file ... Permission denied`.
- Manual cleanup of stale extracts without `sudo`.

## Why `ors_graphs/` doesn't have this problem

`ors_up_cli.py` already has the right pattern — for the **other** bind-mounted directory:

```python
state_graphs_dir = os.path.join(ORS_GRAPHS_DIR, f'{args.state}-buffered')
os.makedirs(state_graphs_dir, exist_ok=True)
```

run **before** `docker compose up`, from the host-only script, specifically "so the bind mount resolves to a host-owned (not root-owned) dir" (per the existing docstring). Docker uses an already-existing host directory as-is rather than auto-creating one, so this directory stays host-owned. **`datasets/openrouteservice/` never got the equivalent treatment** — the code only checks that the buffered `.pbf` file inside it exists, never pre-creates the parent directory itself.

## Suggested fix

Mirror the existing `os.makedirs(..., exist_ok=True)` pattern for `datasets/openrouteservice/` — likely in `ors_up_cli.py` alongside the `ors_graphs` pre-creation, or in `build_buffered_extract_cli` (whichever runs first in the pipeline) — before any `docker compose up` call touches that path.

## Context / how this was found

While merging #319 (adds `.gitignore` rules + `.gitkeep` scaffolding for these directories on `dev`) into a branch where `datasets/openrouteservice/` had already been created root-owned by a prior ORS run, `git pull` failed and needed manual `sudo`/directory-rename workarounds to resolve. That's a one-time hit for anyone who ran ORS before this fix lands, but will recur for **every** teammate who runs the ORS workflow fresh until this is patched.

## Minor doc gap (optional, same ticket or separate)

`CLAUDE.md`'s "DevContainer" section says the project is "designed to be developed inside `.devcontainer/`" without flagging that `ors_up_cli`/`ors_down_cli` are a deliberate host-only exception (that detail only lives in `docs/to_run.md`). Worth a one-line pointer once this branch merges to `dev`, for discoverability.

abd1tus:
# feat(ors): driving-distance tooling + cross-border buffered extract

## 1. Purpose

This PR merges the `feature/driving-distance-tools` epic into `dev`.

It adds the tools that make a driving-distance CSV for a county:

- the OpenRouteService (ORS) container and its lifecycle commands;
- the CLI that routes every census block to every potential polling location;
- a buffered state extract, so routes near a state border can use roads in the next state.

The `feature/driving-time-metric` branch is built on this epic. Merge this PR first. That
freezes this code and makes the driving-time PR smaller.

## 2. How to merge

CAUTION: Merge with a merge commit. Do not squash.

`feature/driving-time-metric` was branched from this epic and carries the first part of
its commit history. A squash gives those commits new hashes, and the driving-time PR then
conflicts with itself.

## 3. What this PR contains

1. **ORS lifecycle.** `ors_up_cli` and `ors_down_cli` run on the host. They start and stop
   the ORS container from `docker-compose.ors.yml` (ORS v8.0.0, pinned). Each state has its
   own graph cache. `run.py` runs these two commands on the host, outside the app container.
2. **Driving-distance matrix.** `ors_url`, `ors_client`, and `driving_distance_matrix` in
   `python/utils/`. The matrix call is batched: at most 2,500 cells and 10 sources per
   request. When a batch fails, each source in it is retried one pair at a time through
   the directions endpoint. A pair with no route is dropped. A negative distance stops
   the run with an error.
3. **`generate_driving_distances_cli`.** Reads the config, the census block centroids, and
   the potential-locations CSV from disk. Writes the three-column CSV the solver reads. A
   rerun fetches only the pairs that are not in the file yet. An origin with no route
   stops the run: the CLI writes the partial CSV, lists each missing origin with its
   coordinates, and exits with code 1.
4. **Buffered extract, the #226 fix.** ORS routes only on the roads it is given. A
   single-state map stops at the state line, so a block near the border gets a longer
   in-state detour that ORS reports as a valid distance. `build_buffered_extract_cli`
   cuts the state plus a 50 km margin out of a cached full-US OSM file, using the state
   outline in `datasets/boundaries/us_states.geojson`. ORS then builds its graph from
   that file and finds the cross-border routes. The orchestrator runs this step
   automatically and skips it when the file exists.
5. **Detour-ratio diagnostic, #226 phase 1.** `analyze_detour_ratios_cli` divides each
   pair's driving distance by its straight-line (haversine) distance and prints the
   distribution per county: median, 90th, 95th and 99th percentiles, and counts over
   1.5, 2.0 and 3.0. A fat upper tail is the fingerprint of missing cross-border roads.
   It reads the solver-side distance files and changes nothing. It was the tool used to
   size the border problem while this PR was developed, and to confirm the September
   Tarrant regeneration against the May file. It is manual today. Running the check
   automatically at the end of every distance run, with a warning on a heavy tail, is
   #362, to keep this PR's scope where the July review left it.
6. **Docs.** `docs/to_run.md` has a section on generating driving distances, including
   the procedure for unrouted origins. `python/CLAUDE.md` gains three comment rules
   (signposts, misattributed "why", test docstrings) and two rules for deleting tests.
7. **Tests.** 15 new test files. One test, `ors_live_buffer_test.py`, needs a live ORS
   with a West Virginia buffered graph and is skipped without it.

## 4. Changes since the July 25 review

Three PRs landed after the first review. #351 and #352 merged into the #337 branch, and
#337 merged into this epic on 2026-09-13.

- The haversine "snap" for unroutable origins is removed. An unrouted origin now stops
  the run (see item 3 above).
- Negative distances now raise at generation time. They are no longer coerced to NaN.
- `resume_from_partial_output` is renamed `identify_unmatched_pairs`.
- `--check-bad-locations` is removed.
- `--state` is replaced by the `--testing` flag. The state comes from the config's
  location.
- The seven inline comments from the July review are answered and resolved on this PR.

## 5. Tickets

Merging into `dev` does not close tickets by itself. Close these by hand after the merge:

- #216 (the epic), #217, #219, #220, #223, #224: done in this PR.
- #225 (negative-distance guards): done, in a stricter form than the ticket asked for.
  The generation-side guard raises instead of coercing to null.
- #218 (auto-detect the `.pbf` filename): the goal is met by a different route. The
  compose file selects the file from the state name, so switching states needs no
  file edit.
- #305 (resume never resumed), #320 (root-owned ORS directory): fixed in this PR.

#226 (the cross-border fix) closes with this merge. Its acceptance test, the `ors_live`
test in item 7, was run against a live West Virginia buffered graph on 2026-08-16 and
passed: the route crossed north of the WV/PA border. The record is on #226.

Follow-ups, already filed: #336 (driving-time reconciliation), #338 (unroutable-origin
walkthrough), #359 (incremental update), #360 (`-l` to `-c`), #361 (DB source for
potential locations), #362 (automatic detour-ratio warning).

## 6. Verification

Run from the epic checkout inside the container, against a test dataset migrated from
`dev`:

- `pytest`: 368 passed, 1 skipped (the `ors_live` test).
- `pylint` on the 31 Python files this PR adds or changes: 10.00/10.
- `pylint python/` overall reports findings in census and database files that this PR
  does not touch. The same findings are on `dev` today.

## 7. Open items, not blocking

- **Decision record for the buffered extract.** This PR cuts each state's road network
  out of the full-US OSM file with a 50 km buffer past the state line, instead of using
  the plain single-state download. Reason: a single-state graph cannot route through
  the neighboring state, so a block near the border gets a long in-state detour that
  looks valid and is wrong. That reasoning, and the options that were rejected, are on
  #226. It is not yet in `docs/development/decisions/` as an Architecture Decision
  Record (ADR).
  Write the ADR after this merges.
- **Credit to `dxwils3/geolib`.** The driving-distance code here is a reimplementation
  in this project's layered structure, not a copy. A courtesy line in the module
  docstrings pointing at the upstream project would be good practice.

https://claude.ai/code/session_01PxcVzp78SCHXu2i8eAdgWe


python/scripts/analyze_detour_ratios_cli.py
```diff
@@ -0,0 +1,79 @@
...
+
+def _build_arg_parser() -> argparse.ArgumentParser:
+    """Build argument parser for the CLI.
+
+    Returns:
+        Configured ArgumentParser instance.
+    """
+    parser = argparse.ArgumentParser(
+        prog='analyze_detour_ratios_cli',
+        description='Report per-county driving/haversine detour-ratio distribution (#226 diagnostic).',
```
Amasus, line 30: Take the ticket references out. It will just confuse a future reader.

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+        help='Probe-only mode: list unroutable origins and exit without writing output.',
+    )
+    parser.add_argument(
+        '-v', '--verbose', action='count', default=0,
+        help='Increase screen verbosity (-v or -vv). The log file captures everything regardless.',
+    )
+    return parser
+
+
+def build_output_csv_path(location: str) -> str:
```
Amasus, line 75: Repeated code. See python/utils/utils.py:139

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+    for _, row in pots_df.iterrows():
+        loc_id = str(row['Location'])
+        lon, lat = extract_lon_lat(row)
+        locations[loc_id] = [lon, lat]
+        dest_ids.append(loc_id)
+
+    return locations, source_ids, dest_ids
+
+
+def _pick_coord_extractor(columns):
```
Amasus, line 117: This is (or should be dead code). The documentation states that "lat, lon" is the correct column. The code ignores everything else. There is a remaining "latitude" and "longitude" pair that exists in the test fixture. This should be removed. See ticket #324. 

Also, this function would completely miss a check of a file coming from the data base, where the column is called lat_lon

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+        config: A loaded ``PollingModelConfig``.
+
+    Returns:
+        A tuple ``(locations, source_ids, dest_ids)``:
+            - ``locations``: ``id -> [longitude, latitude]`` dict.
+            - ``source_ids``: census block GEOIDs (origins).
+            - ``dest_ids``: potential-location ``Location`` values (destinations).
+    '''
+    blocks_gdf = get_blocks_gdf(config.census_year, config.location)
+    pots_df = load_potential_locations_csv(
```
Amasus, line 95: This is assuming that the potential_locations file is stored locally. Are the plans to generalize this so that it can read from the database as well? e.g. if I want to run driving distances for a county where I've already done haversine distance?

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+    '''Return an open log file handle for this run, plus its path.'''
+    os.makedirs(logdir, exist_ok=True)
+    basename = os.path.basename(config_file_path)
+    log_path = os.path.join(logdir, f'{log_date_prefix()}_{basename}_driving_distances.log')
+    # Caller closes the handle in a try/finally; deliberately not a context manager.
+    log_fh = open(log_path, 'a', encoding='utf-8')  # pylint: disable=consider-using-with
+    return log_fh, log_path
+
+
+def _tee(message: str, log_fh, *, to_screen: bool = True) -> None:
```
Amasus, line 177: rename to _log_and_print , or similar for readability reasons?

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+            ``/matrix/...`` and appending ``/health``.
+    '''
+    health_url = matrix_url.rsplit('/matrix/', 1)[0] + '/health'
+    try:
+        with urllib.request.urlopen(health_url, timeout=5) as response:
+            if response.getcode() == 200:
+                return
+    except (urllib.error.URLError, ConnectionError, TimeoutError):
+        pass
+    print(
```
Amasus, line 199: can we clean up the extra f's here. it just makes it hard for a human to read.

python/scripts/generate_driving_distances_cli.py
```diff
@@ -0,0 +1,323 @@
...
+
+    config = PollingModelConfig.load_config(args.location_config)
+
+    state = args.state
+    if state is None:
+        try:
+            state = state_slug_from_location(config.location)
+        except ValueError as exc:
+            print(
+                f'Couldn\'t derive state from config (location={config.location!r}; {exc}).\n'
```
Amasus, line 250: Two things: 
1. This seems to be duplicated error messaging. see run.py 401.
2. Can we change the error message to something more precise: E.g. `Couldn't` derive a state from location={location!r} ({exc}).

If you are running a non-geographic config (e.g. location: testing)
pass --state explicitly (e.g. GA for testing)

Otherwise: Fix your `location` field. Do not pass a separate state. `

python/scripts/analyze_detour_ratios_cli.py
abd1tus: Done in fd891ecd

python/scripts/generate_driving_distances_cli.py
abd1tus: Done in 5e6c2753. `build_output_csv_path` has been removed.

python/scripts/generate_driving_distances_cli.py
abd1tus: Resolved by #351. The function removed.

python/scripts/generate_driving_distances_cli.py
abd1tus: CVS only still.  I created a ticket #361 so that we can address potential location pulls from db in a future enhancement.

python/scripts/generate_driving_distances_cli.py
abd1tus: Done in ea5650c0.

python/scripts/generate_driving_distances_cli.py
abd1tus: Done in a55d7cbe.

python/scripts/generate_driving_distances_cli.py
abd1tus: Resolved by #352, your PR. `--state` is gone and `--testing` replaced it.

Amasus:
### Overview

Changes made since last review look fine. Thank you for changing the _tee function name. A few outstanding notes:
1. I had Claude make a sweep to find references to tickets (grep #[0-9]+ essentially). It's created a table for you below.
2. Yes to crediting Dan. Good catch
3. Agreed. Writing an ADR is a good idea. #226 has the meat of it. Better to have it in an ADR than a closed ticket.
4. I caveat, I've moved the relevant tickets to "prestaged on feature branch". Once this is merged, they should be moved to staged on dev, not done just yet.

### Leftover ticket references on this branch

Following the two cleanup commits (`fd891ecd`, `4fe6e518`), I swept the whole branch for ticket references. There are **17 sites across 11 files**. All of them are introduced by this branch — of these files only `docs/to_run.md` exists on `dev`, and it carries no ticket references there.

Splitting them into the ones that look like cleanup candidates and the ones that should probably stay:

#### Provenance references — 13 sites, 9 files -- please remove these

These say "this exists because of #N", which is what the two cleanup commits were removing.

| File | Line(s) | Ticket |
|---|---|---|
| `.devcontainer/docker-compose.ors.yml` | 15 | #223 — "graph from a previous state is silently reused (see issue #223)" |
| `python/tests/run_py_test.py` | 145, 165 | #223 — "(#223)", "this regression test guards #223 fix" |
| `python/scripts/build_buffered_extract_cli.py` | 1, 26 | #226 — module docstring and argparse `description` |
| `python/utils/buffered_extract.py` | 1 | #226 — module docstring |
| `python/tests/ors_live_buffer_test.py` | 1, 4, 11 | #226 — docstring, "#226 restores…", plus "The ticket's WV->PA pair" on line 4 (worded, no number, so it won't show up in a `#[0-9]+` grep) |
| `python/scripts/ors_up_cli.py` | 221 | #320 — "See #320." |
| `python/tests/ors_up_cli_test.py` | 185 | #320 — "See #320." |
| `python/tests/driving_distance_matrix_test.py` | 316 | #294 — "A negative distance_m fails loudly at generation (#294)" |
| `python/utils/driving_distance_matrix.py` | 334 | #305 — "…never match the str source/dest ids. See #305." |

#### Forward-looking TODOs — 4 sites, 3 files — Don't remove these. Just for tracking.

| File | Line(s) | Ticket |
|---|---|---|
| `docs/to_run.md` | 138 | #338 — "TODO: update after #338 is implemented." |
| `python/scripts/generate_driving_distances_cli.py` | 191, 197 | #338 — "see #338", "TODO: see #338." |
| `python/utils/driving_distance_matrix.py` | 258 | #338 — "see #338" |

#338 is still open and unassigned, so these point at real outstanding work rather than at history. They read as accurate to me as written.

#### Two spots worth a closer look

1. **`driving_distance_matrix_test.py:316` is probably a miss.** `4fe6e518` dropped the ticket numbers from the negative-distance docstring but touched only `python/utils/driving_distance_matrix.py`. The test docstring for that same behavior still cites #294.

5. **`driving_distance_matrix.py` was cleaned but still carries two references** — #305 at line 334 and #338 at line 258. The cleanup has been per-file rather than a sweep, so it is worth one pass over the whole branch rather than file-by-file.

Not a ticket reference, so excluded above: `.gitignore:107` mentions `pypa/pipenv#598`, which is upstream boilerplate from the standard Python gitignore template.


### Follow-up tickets — can we talk these through?

I've assigned all six follow-ups from section 5 to you. I'd love to get an understanding of what the status of these are, update that on the board, and track them properly. If I need to take some of them on, I can do that as well.

| Ticket | Board status |
|---|---|
| #336 driving-time reconciliation | Backlog |
| #338 unroutable-origin walkthrough | Backlog |
| #359 incremental update | Icebox |
| #360 `-l` to `-c` rename | Icebox |
| #361 DB source for potential locations | Icebox |
| #362 automatic detour-ratio warning | For discussion |

- Note that #338 is referenced by live TODOs in `generate_driving_distances_cli.py`, `driving_distance_matrix.py`, and `docs/to_run.md`.

Amasus:
Per discussion on 9/21, Chad will look at #338 after this is merged, and before touching drive time work. The rest of the follow up tickets are low priority. They have all been assigned statuses

PR #312, opened 2026-07-14
feat(precincts): add solver-assignment drive-time heat maps with shared color scale

Amasus:
## Summary
- Adds a parallel set of drive-time heat maps sourced from the solver's own results (`OPTIMIZATON_RESULTS`, the existing-locations-only optimization run), alongside the existing precinct-crosswalk-based maps — `flagged_optimized_distant_blocks()` reshapes the results CSV into the same schema `make_demo_distance_heat_map()` already consumes, and re-adds the zero-population blocks the optimizer excludes (verified: 2772 total blocks - 701 zero-population = 2071, matching the results CSV row count exactly) as gray placeholder rows.
- `make_demo_distance_heat_map()` gains `map_label` (distinguishes filenames/titles between the precinct-assignment and solver-assignment maps) and `color_bounds` (a shared min/max duration range applied via `scale_fill_gradient()`/`scale_color_gradient()` `limits=`) so all 8 heat maps use one apples-to-apples color scale instead of each auto-scaling independently.
- Replaces the 5-mile distance-proxy flagging with real drive-time (`duration_min`) thresholds at 15/20 minutes (carried over from the prior commit on this branch, closing #271).

## Review notes

Author has concerns about code quality. This was done quick and dirty. Please review with that in mind.

## Test plan
- [x] `Rscript R/result_analysis/extract_precincts.r` runs to completion with no errors, produces all 8 heat map PNGs plus the 4 intermediate CSVs
- [x] Verified backfilled zero-population blocks: both `optimized_distance_flagged_blocks_{15,20}_min.csv` have 2772 rows (full block count), matching `block_precinct_assignment`
- [x] Recomputed the shared color scale independently from the written CSVs (min ≈15.0, max ≈53.4 min) and confirmed it matches `duration_color_bounds`
- [x] `lintr::lint()` on touched files shows only pre-existing warnings (NSE column bindings, line-length style) in unrelated code, nothing new introduced by this diff
- [ ] Visual review of the 8 PNGs for correct color-scale/legend rendering

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #323, opened 2026-07-24
fix(precincts): add polling-location dots, solver-assignment precinct outlines, and spreadsheet-safe GEOID export

Amasus:
## ⚠️ Review notes

- **This code has not yet had a human review — it is almost entirely AI-written** (paired with Claude Code). Please review with that in mind.
- **This branch is built off `feature/optimization_output_maps` (#312), not `delivery/Monongalia_County` directly.** Merge order matters:
  - If this PR merges into `feature/optimization_output_maps` **before** #312 merges into `delivery/Monongalia_County`, everything lands together in one merge — no extra step needed.
  - If #312 merges into `delivery/Monongalia_County` **first**, `feature/optimization_output_maps` will still have this PR's commit sitting on top, unmerged into delivery — a follow-up PR (or cherry-pick of this commit directly onto `delivery/Monongalia_County`) will be needed to actually get these client-requested tweaks to the client-facing branch.
  - Recommend merging this one first if both are close to ready, to avoid the extra follow-up step.

## Summary
Client-requested tweaks to the Monongalia County heat maps, built on top of
feature/optimization_output_maps (#312).

- Add polling-location point markers to all block-level distance heat maps
  (both actual-assignment and solver-assignment), reusing the existing
  Okabe-Ito "polling" blue/shape from tableau_theme.R for visual consistency
  with the county's other reference maps. Colorblind-safety checked against
  the existing red duration gradient (ΔE 26-38 across protan/deutan/tritan,
  well above the safety threshold).
- Step 5's solver-assignment heat maps now draw the solver's own dissolved
  precinct outlines (from Basic_analysis.r's make_precinct_map() output)
  instead of the state-provided precinct shapefile, since these maps
  represent the solver's grouping, not the as-provided precincts. Reading
  that shapefile fails loudly (missing file, or older than the results CSV
  it should reflect) instead of silently plotting stale boundaries.
- Fix distance_flagged_blocks_*_min.csv / optimized_distance_flagged_blocks_*_min.csv
  so their id_orig (census block GEOID) column opens as text in Excel/Sheets
  instead of being silently reinterpreted as a number (which would drop
  leading zeros for any state whose FIPS code starts with 0). Wrapped as an
  Excel/Sheets formula-text literal before writing; in-memory objects used
  for the heat-map merges are unaffected.
- Regenerated all affected heat map PNGs and block-level CSVs for
  Monongalia_County_WV.

## Test plan
- [ ] Rerun `extract_precincts.r` for Monongalia_County_WV and visually
      confirm: polling dots appear on all 8 heat maps, Step 5 maps show the
      solver-dissolved precinct outlines, Step 4 maps still show the
      as-provided state precincts.
- [ ] Open distance_flagged_blocks_15_min.csv in Excel/Sheets and confirm
      id_orig displays as full digit strings, not scientific notation.
- (R analysis tooling has no automated test suite per R/CLAUDE.md.)


Amasus:
One note: given how empty precincts are assigned to locations (after PR #312) , does it make sense to change how the precinct maps are used for the solver defined precinct denotation here?

PR #326, opened 2026-07-27
feat(scripts): fetch ORS routes for Monongalia's long-drive-time blocks

Amasus:
## Review notes

- Running `fetch_long_drive_time_routes.py` requires the ORS sibling container up (`python3 run.py ors_up_cli west-virginia`, host-only — can't be run from inside the devcontainer).

## Summary

Client deliverable: the actual road routes for the 8 populated census
blocks over the 30-minute drive-time threshold in
`distance_flagged_blocks_20_min.csv`, for a follow-up client discussion about those specific blocks.

- `python/scripts/delivery_scripts/fetch_long_drive_time_routes.py` — fetches
  each route via ORS's single-pair directions endpoint, reusing the same
  origin/destination coordinates `generate_driving_distances_cli.py` used to
  build the original distance matrix (via its existing
  `derive_origins_and_destinations()`). Retries with a wider snap-to-road
  radius when the directions endpoint's stricter default rejects a point the
  matrix endpoint already routed successfully (confirmed empirically: one
  block's centroid sits 531.91m from the nearest mapped road — the matrix
  service tolerates that by default, the single-pair directions service
  doesn't unless asked to). Raises loudly rather than silently dropping a
  row if a route still can't be found for a block that's already known to be
  routable, since that would indicate something is genuinely wrong rather
  than an expected edge case.
- `precinct_analysis_outputs/Monongalia_County_WV/long_drive_time_routes.geojson`
  — the 8 fetched routes, going to the client.
- `python/CLAUDE.md` — documents the new `python/scripts/delivery_scripts/`
  convention: one-off/narrowly-scoped scripts that produce a specific
  deliverable, committed alongside that deliverable for traceability,
  following the same TDD conventions as the rest of `python/`.

## Test plan
- [ ] Spot-check a couple of the 8 routes in the geojson against the ORS
      graph (e.g. via QGIS) to confirm they look like plausible driving
      paths, not snapping artifacts.
- (One-off delivery script per the new `python/CLAUDE.md` convention — TDD
  coverage follows that convention, not a blanket exemption.)


antisocialscientist (changes requested):
My only concern here is that id_dest isn't forced to be a string. If the county has numbered precincts then couldn't id_origin and id_dest end up as different types and much hilarity could ensue?

Issue #249, opened 2026-06-18
[RDH-pop] Documentation


Amasus:
Part of #243. Blocked on #244.

Update `docs/input_files.md` with a section parallel to the CVAP section. Explicitly explain why we use this RDH dataset instead of pulling ACS data directly: ACS is only available at the block group level, while this RDH dataset is disaggregated to the block level, which the solver requires. Update `data_dictionary.md` accordingly.

PR #331, opened 2026-07-31
docs,refactor(datasets): document RDH predicted VAP and rename to projected_vap

DashaU3:
#249

## Summary
- `docs/input_files.md`, `docs/data_dictionary.md`, `docs/references.md`, `docs/how_to_cite.md`, `docs/software.md`: document why RDH's own projected VAP dataset is used instead of pulling ACS data directly (ACS publishes no future-year projections at any geography; RDH builds its own projection from block-level census and block-group ACS data), add citation/terms-of-use parity with the CVAP dataset, and add the `census_data_type`/`projection_year` columns to the `model_configs` table in the data dictionary.
- Renames `predicted_vap`/`predicted VAP` to `projected_vap`/`projected VAP` throughout the codebase: the `census_data_type` config value, the `RDH_predicted_vap` -> `RDH_projected_vap` dataset directory, Python identifiers (functions, constants, tests) in `python/solver/`, `python/database/`, and `python/utils/`, and committed test fixtures. No behavior change; existing configs or DB rows using the old value are not migrated.

## Test plan
- [x] `pylint python/` — unchanged score (9.70/10), no new warnings
- [x] `pytest python/tests/ -m "not e2e_db"` — 244 passed


docs/data_dictionary.md
Amasus, line 1: Ai review of this file:

Two specific things are off here, and they're classic LLM tells rather than just "different from your voice":

Explicit causal signposting ("too coarse for this model, which..." / "which is why...") — the doc otherwise just states what a field is, not the chain of reasoning for why the pipeline uses it that way. A data dictionary records facts; it doesn't argue for design decisions.
A self-referential backward pointer ("the same role it plays for CVAP data above") — this is the model trying to be helpful by drawing an explicit connection for the reader, but nowhere else in the doc does a section reference another section this way. It reads like a tutor walking you through logic, not a reference entry.

Also worth noting: it re-explains something the whole document already assumes ("this model, which assigns individual census blocks to polling locations") — restating context that's established on line one of the file. That kind of belt-and-suspenders context-restating is a very common LLM habit.

My suggestion: keep bullets 1, 3, 4 as-is (they're fine, consistent with your style), and cut bullet 2 down to a flat factual statement, e.g. something like: "Census/ACS projections are only published at the block group level; Redistricting Data Hub disaggregates them to the block level, same as for CVAP data." One sentence, no "which is why," no reasoning chain — just the fact, in your document's register.

docs/data_dictionary.md
```diff
@@ -2,7 +2,7 @@
 
 This document provides column-level definitions for all database tables and views in the `equitable_polling_locations_prod` BigQuery dataset. For an overview of the database, see [database](database.md).
 
-Demographic population fields in this dataset are sourced either from the U.S. Census Bureau's Decennial Census redistricting data, or from Redistricting Data Hub's disaggregation of the ACD CVAP (citizen voting age data) Specifically, for the redistricting data:
+Demographic population fields in this dataset are sourced from one of three places, selected per-config by `census_data_type`: the U.S. Census Bureau's Decennial Census redistricting data, Redistricting Data Hub's disaggregation of ACS CVAP (citizen voting age population) data, or Redistricting Data Hub's disaggregation of projected VAP (voting age population) data. Specifically, for the redistricting data:
```
Amasus: Tiny nit, this would t
Read better if the parenthetical clause was its own sentence.

docs/input_files.md
```diff
@@ -113,6 +113,8 @@ Stored in `datasets/census/RDH_predicted_vap/<County_ST>/`
 
 RDH predicted VAP data provides block-level population projections for the Voting Age Population (VAP) sourced from the [Redistricting Data Hub](https://redistrictingdatahub.org/). These projections are based on the 2020 census (P3/P4 tables) and cover multiple future years in a single file (e.g., 2026–2035). The `projection_year` config field selects which year's columns are used at model-run time.
 
+**Why RDH instead of pulling ACS data directly:** Census/ACS population projections are only published at the block group level. The solver assigns individual census blocks — not block groups — to polling locations, so block-group-level projections are too coarse to use directly. Redistricting Data Hub disaggregates these projections down to the block level, the same disaggregation role it plays for CVAP data above, which is why this model uses RDH's predicted VAP dataset rather than pulling ACS projection data directly.
```
Amasus: This explains why not acs. This makes it sound like we should be using RDH ACS disaggregation data. But we're not. We are using a projection model. Why?

Amasus:
Is there a place where you point to the rdh documentation? Also, did you include the projection data in the set of things that needs a citation or agreement to terms and conditions? Alternatively, did you add the terms and conditions/ citation needs where you tak about this data source? Parallel to how it is done for cvap? ( I forget the direction I organized things last time).

DashaU3:
> Is there a place where you point to the rdh documentation? Also, did you include the projection data in the set of things that needs a citation or agreement to terms and conditions? Alternatively, did you add the terms and conditions/ citation needs where you tak about this data source? Parallel to how it is done for cvap? ( I forget the direction I organized things last time).

I didn't do the citations previously, but I've now added them parallel to how it's done for cvap.

docs/input_files.md
DashaU3: This was an error due to paralleling the explanation for CVAP too closely, but in reality, RDH processes the two types of data separately (it creates its own projections, since ACS doesn't do that), so it's been updated to be accurate for VAP data.

Amasus (commented):
Sorry, doing this from my phone. These comment didn't make it in.

docs/input_files.md
```diff
@@ -113,6 +113,10 @@ Stored in `datasets/census/RDH_predicted_vap/<County_ST>/`
 
 RDH predicted VAP data provides block-level population projections for the Voting Age Population (VAP) sourced from the [Redistricting Data Hub](https://redistrictingdatahub.org/). These projections are based on the 2020 census (P3/P4 tables) and cover multiple future years in a single file (e.g., 2026–2035). The `projection_year` config field selects which year's columns are used at model-run time.
 
+**Why RDH instead of pulling ACS data directly:** This isn't the same situation as CVAP, where ACS already publishes a block-group-level product that RDH disaggregates. The Census Bureau/ACS does not publish future-year population projections at all, at any geography — projections require a forecasting methodology, not just tabulation of a survey. RDH produces the projection itself: it uses the 2020 decennial census and 5-year ACS estimates only as historical inputs to its own modeling, then delivers the result at the census block level, the granularity this model requires. There is no ACS projection product to pull directly instead.
```
Amasus: Please read the LLM stuff before you send it to me. Try to put it into a reference text or standard readme voice. Run it by a different human if you need.

docs/data_dictionary.md
```diff
@@ -2,7 +2,7 @@
 
 This document provides column-level definitions for all database tables and views in the `equitable_polling_locations_prod` BigQuery dataset. For an overview of the database, see [database](database.md).
 
-Demographic population fields in this dataset are sourced either from the U.S. Census Bureau's Decennial Census redistricting data, or from Redistricting Data Hub's disaggregation of the ACD CVAP (citizen voting age data) Specifically, for the redistricting data:
+Demographic population fields in this dataset are sourced from one of three places, selected per-config by `census_data_type`: the U.S. Census Bureau's Decennial Census redistricting data, Redistricting Data Hub's disaggregation of ACS CVAP (citizen voting age population) data, or Redistricting Data Hub's own projected VAP (voting age population) modeling. The projected VAP modeling is built from decennial census and ACS data as historical inputs. Specifically, for the redistricting data:
```
Amasus, line 5: this one looks good. Thanks.

docs/data_dictionary.md
```diff
@@ -12,6 +12,12 @@ For the CVAP data:
    - Note that race and ethnicity are calculated differently in these tables than for redistricting
 - These population counts represent the *citizen* voting-age population (18+) at the census block level.
 - All uses of data products based off of Redistricting Data Hub's data must comply with their [terms and conditions](https://redistrictingdatahub.org/terms-and-conditions/).
+
+For the RDH predicted VAP data (`census_data_type: predicted_vap`):
+- Population projections are based on the 2020 census (P3/P4 tables) and cover multiple future years in a single file; the config's `projection_year` field selects which year's columns are used. See Redistricting Data Hub's [projections methodology page](https://redistrictingdatahub.org/projections/).
+- Unlike CVAP, this isn't a disaggregation of an existing ACS product — the Census Bureau/ACS does not publish future-year population projections at all, at any geography. RDH produces the projection itself, using the 2020 decennial census and 5-year ACS estimates only as historical inputs, and delivers the result at the block level this model requires.
```
Amasus: I still don't like this. Please review yourself and make certain it is consistent with the rest of the readme's style.

docs/data_dictionary.md
Amasus: in retrospect, as a human reading your text, i really think it needs to be read carefully by a human and not an LLM. I don't like bullet 1 either. Bullet 2 is still off. Not sure why 3 is off by it's lonesome. Why is it using predicted instead of projected? if you need you explain what you mean by projected, do so. Don't just put in a synonym.

Amasus (commented):
please actually review the text you submit and make certain it matches style, and is suitable for a clear readme document.

docs/input_files.md
```diff
@@ -111,7 +111,9 @@ Note, any use of data or data products built on data from Redistricting Data Hub
 
 Stored in `datasets/census/RDH_predicted_vap/<County_ST>/`
 
-RDH predicted VAP data provides block-level population projections for the Voting Age Population (VAP) sourced from the [Redistricting Data Hub](https://redistrictingdatahub.org/). These projections are based on the 2020 census (P3/P4 tables) and cover multiple future years in a single file (e.g., 2026–2035). The `projection_year` config field selects which year's columns are used at model-run time.
+VAP stands for voting age population. This is from the 2020 census. [Redistricting Data Hub](https://redistrictingdatahub.org/) creates projections from 2026 - 2035, as the ACS does not publish projections.
```
Amasus: keep the old text. Not the new one here.

docs/data_dictionary.md
```diff
@@ -12,6 +12,13 @@ For the CVAP data:
    - Note that race and ethnicity are calculated differently in these tables than for redistricting
 - These population counts represent the *citizen* voting-age population (18+) at the census block level.
 - All uses of data products based off of Redistricting Data Hub's data must comply with their [terms and conditions](https://redistrictingdatahub.org/terms-and-conditions/).
+
+For the RDH predicted VAP data (`census_data_type: predicted_vap`):
+- **Race and Ethnicity** come from the 2020 census' P3/P4 tables. See Redistricting Data Hub's [projections methodology page](https://redistrictingdatahub.org/projections/).
+- Note that the dataset used is RDH's own, as ACS data is only available at the block group level, not the block level, and does not include projections at all.
+- Projections are from 2026 - 2035
+- These population counts represent the projected voting-age population (18+) at the census block level.
```
Amasus: Replace above 3 bullets with:
- Note that the dataset used is RDH's own, not a census derivative.
   - The ACS data is only available at the block group level, not the block level, and only for certain years.
   - Instead, RDH develops their [own methodology](link documentation here) using a combination of block level census data and block group level ACS data to project VAP populations from 20236-2035.
   - Note, these are projected values only, not actual values.

Amasus (commented):
There seems to be a consistency issue between projected vap (language in RDH) and predicted VAP (language in our code base. Everything should probably be "projected VAP" in order to avoid confusion

DashaU3:
> There seems to be a consistency issue between projected vap (language in RDH) and predicted VAP (language in our code base. Everything should probably be "projected VAP" in order to avoid confusion

The problem is that when I was first creating the variable, I called it predicted_vap, so now it's referred to as such throughout the codebase. Would it potentially cause problems to change it now, or is it safe to do so? Also, if I do change it, should I add that to this PR, or somewhere else (since it goes beyond the scope of just documentation)?

Issue #336, opened 2026-08-16
[driving-time] Reconcile against fix/325: drop diverged estimate_origin (duration_s mismatch) and adopt the #305 resume dtype fix


abd1tus:
## Summary

`feature/driving-time-metric` (#227) carries diverged copies of the unrouted-origin machinery that `fix/325-unrouted-origin-handling` removed on the driving-distance side. Once fix/325 merges into the `feature/driving-distance-tools` epic and the epic flows into driving-time, that branch must be reconciled: rebasing/merging it onto the reworked code **drops its diverged copies**, which fixes two defects at once.

This is the follow-on ticket mandated by #325 item 6's "**Depends on / blocks `feature/driving-time-metric` (#227)** — needs a follow-on ticket, not resolved here" paragraph.

## The two defects on `feature/driving-time-metric`

1. **Fabrication + distance/time inconsistency in its `estimate_origin`.** The driving-time copy of `estimate_origin` / `_snap_unroutable_origins` carries a snapped row's `duration_s` **unchanged** while correcting `distance_m` by the haversine offset (explicitly, because converting that offset to a travel time would require an invented speed constant). Beyond being the same silent fabrication #325 item 6 removed, this creates a real inconsistency between the two metrics on the same row. Removal via reconciliation fixes both: no snapped rows means no fabricated distances *and* no mismatched durations.

2. **Missing #305 dtype fix.** Its `resume_from_partial_output` still does a plain `pd.read_csv(output_path)` without `dtype=str` on the id columns, so numeric-looking GEOIDs load as int64 and the resume present-pair check never matches — the bug fixed on the ddt side in `45a1885c` (resume re-fetches everything and duplicates rows via mismatched dedup keys).

## What reconciliation looks like

Blocked by, in order:
1. `fix/325-unrouted-origin-handling` → `feature/driving-distance-tools` (PR forthcoming)
2. epic sync into `feature/driving-time-metric`

Then, on driving-time:
- **Drop** its `estimate_origin` / `_snap_unroutable_origins` / `HAVERSINE_SNAP_RADIUS_METERS` and their tests in favor of the epic's drop-and-fail-loud behavior (`build_distance_matrix` returns only routed rows; the CLI exits non-zero listing missing origins with id + lat/lon; partial CSV is kept for resume).
[...]

Issue #325, opened 2026-07-25
generate_driving_distances_cli: dropped/unrouted origins are silent, undocumented, and validated too late


Amasus:
## Summary

`generate_driving_distances_cli.py` and the driving-distance-matrix pipeline it calls into can silently drop origins that ORS cannot route and has no in-range haversine neighbor to snap to (`_snap_unroutable_origins`, `python/utils/driving_distance_matrix.py:271-326`). Neither the diagnostic `--check-bad-locations` mode nor the normal run path treats this as a failure — both exit 0 regardless. The only enforcement point in the whole pipeline is much later, at model-run time (`python/solver/model_data.py:751-775`), decoupled in time, person, and context from wherever the CSV was generated. This ticket tracks making unrouted-origin handling explicit, documented, and validated at generation time, with a human in the loop rather than automated resolution — including removing the automated haversine-neighbor estimation step entirely (see #6).

Furthermore, `resume_from_partial_output` seems to accidentally protect against the case where potential_locations_files are changed. But this is not documented.

Reviewed as part of PR #321.

### 1. Dropped origins never block progress anywhere in this script

- `--check-bad-locations` mode (`generate_driving_distances_cli.py:269-280`) computes the set of unroutable origins and prints/logs it, but always `return 0` regardless of whether any were found.
- The normal (write) path (`generate_driving_distances_cli.py:282-317`) is worse: it doesn't even print an aggregate summary. The only trace of a dropped origin is a single inline log line emitted during `build_distance_matrix` (`unroutable origin X: no neighbor within 1km, dropped`) — no end-of-run rollup.
- Every populated block needs a real driving distance for the solver to run at all (enforced later at `model_data.py:751-775`). Silently proceeding here just relocates the failure to a confusing, disconnected point later in the pipeline, potentially hit by a different person than whoever generated the file.
- **Needed:** when any origin is dropped, still write the CSV with whatever was successfully computed (don't lose completed work — this also composes correctly with `resume_from_partial_output`, see #3), but exit non-zero and print a clear, actionable message enumerating exactly which origins are missing. This is expected to require human intervention (e.g. manually researching/entering a distance, correcting a bad centroid) rather than automated resolution.
   - **Needed:** Clear instructions of how a human should handle the data fixes. Changes to census data? changes to a temp file? Does this need a helper script? What is the process?

### 2. `--check-bad-locations` has zero documentation. Why is it needed?

[...]

PR #337, opened 2026-08-16
fix: remove fabricated distances, fail loud on unrouted origins (#325)

### LINKED ISSUE #224: Derive ORS state from config location instead of requiring a CLI arg

### LINKED ISSUE #324: testing_potential_locations.csv fixture has redundant Latitude/Longitude columns not in the documented schema
## Problem

`datasets/polling/testing/testing_potential_locations.csv` has columns:

```
Location,Address,Location type,Latitude,Longitude,"Lat, Lon"
```

The documented schema (`docs/input_files.md`, "Columns" section under potential_locations) only requires `Location`, `Address`, `Location type`, and `Lat, Lon` (a single comma-separated string). There is no documented `Latitude`/`Longitude` pair. Real county data (`Tarrant_County_TX`, `Monongalia_County_WV`) only has `Lat, Lon`, matching the documented schema. The testing fixture is the only file with the extra pair, and its values duplicate what's already in `Lat, Lon`.

This is what motivated the `Latitude`/`Longitude` branch in `_pick_coord_extractor()` (`python/scripts/generate_driving_distances_cli.py`), which was written to accommodate the fixture's shape rather than the fixture matching the documented/production schema.

## Fix

Drop the redundant `Latitude`/`Longitude` columns from `testing_potential_locations.csv` so the fixture matches the documented schema and real-world data. Once done, `_pick_coord_extractor`'s separate-columns branch (and its tests) can likely be removed too, since nothing in the codebase would exercise that schema anymore — worth confirming before removing.

## Context

Found during review of PR #321.

abd1tus:
**Base branch: `feature/driving-distance-tools`.**

**CAUTION: Merge this PR with a merge commit. Do not squash.** This branch contains the commits from PR #351 and PR #352. The epic branch must receive these commits without change.

This PR implements the parts of #325 that do not need a decision. PR #351 and PR #352 changed this branch after the first description was written. This description gives the current state of the branch.

## Changes

**1. The CLI stops when an origin has no route.**

- The CLI no longer makes an estimated distance for an origin that ORS cannot route. The functions `estimate_origin` and `_snap_unroutable_origins` are removed.
- The CLI writes the partial CSV file. Then it prints each origin that has no route, with the latitude and longitude. Then it stops with exit code 1.
- A second run gets only the pairs that are not in the CSV file.
- If a CSV file has a row with an empty `distance_m` value, the CLI reports that row in the same way.
- Reference: #325, items 1 and 6.

**2. A negative distance causes an error.**

- The function `_coerce_negative_distances_to_null` is replaced by `_reject_negative_distances`.
- Before this change, the epic branch changed a negative distance to null. After this change, a negative distance stops the run with a `ValueError`.
- Reason: ORS returns null for a route that does not exist. A negative distance shows an error in the routing backend. See #294 and #225.

**3. The option `--check-bad-locations` is removed.**

- The parser for separate `Latitude` and `Longitude` columns is also removed.
- The option gave no advantage over a normal run. No committed data file uses separate columns.
- Reference: #325, items 2 and 7. This change closes #324. Source: PR #351.

**4. The option `--state <slug>` is replaced by `--testing`.**

- A real configuration always gets its state from the `location` field.
- The `--testing` option is only for the `testing` fixture.
- The `run.py` orchestrator sends the `--testing` option to the container. It does not send a state value.
- This change closes #224 and #350. Source: PR #352.

**5. Two functions have new names.**

- `get_missing_origins` is now `get_origins_with_blank_distances`.
- `resume_from_partial_output` is now `identify_unmatched_pairs`.
- The docstrings now describe what each function does.
- The unused `log` parameter of `filter_distance_data` is removed. This change touches `model_data.py`, `model_run.py`, and `conftest.py`. PR #351 reports this change.
- Reference: #325, items 3 and 4.

**6. Two comment rules are added to `python/CLAUDE.md`.** Source: PR #351.

## Test results

- The full test suite: 361 tests pass. 7 tests fail. The same 7 tests fail on `dev`. The causes are the `duration_s` column in the `tests_chad` dataset and the census key in the container environment. This branch does not cause these failures.
- The `testing` fixture on a graph for the wrong state: the CLI writes 0 rows, lists all 10 origins with coordinates, and stops with exit code 1.
- Tarrant County with `run.py` and no option: the CLI gets the state from the configuration, builds the Texas graph, and gets 11,748,400 pairs (31,925 origins × 368 destinations) in 6 minutes. The result has 0 null values, 0 negative values, and 0 origins without a route. The exit code is 0.

## Not in this PR

These items stay open on #325:

- The procedure for a person to correct a block that has no route. See #338.
- The request to redesign the procedure to add locations. See #325, item 3.
- The decision on a potential-locations file that has fewer rows than before. See #325, item 5. There is no decision yet.
- The reconciliation of the driving-time branch. See #336. This merge removes the block on #336.


python/utils/ors_client.py
```diff
@@ -1,7 +1,7 @@
 '''Thin HTTP wrapper around the OpenRouteService matrix and directions endpoints.
 
 This module exposes only the wire-level interactions with ORS. Higher-level
-concerns (batching, retries, snapping) live in driving_distance_matrix.py.
+concerns (batching, retries) live in driving_distance_matrix.py.
```
Amasus, line 4: This is likely a good place for there to be a few bullets on how ORS works and what we are using it for in this project. The current commenting doesn't explain much.

Please make certain that the comments are at the right level of abstraction for a human, not an LLM.

python/utils/driving_distance_matrix.py
```diff
@@ -1,9 +1,10 @@
 '''Orchestrate the building of a driving-distance matrix via ORS.
 
 This module batches ORS matrix calls, retries individual sources via the
-directions endpoint when a batch fails, snaps origins that cannot be routed
-to nearest haversine neighbors, and reshapes the ORS response into the
-project's canonical long-form CSV shape.
+directions endpoint when a batch fails, and reshapes the ORS response into
+the project's canonical long-form CSV shape. Origins ORS cannot route are
+dropped from the result — never estimated — so callers can detect and fail
+loud on them (see #325).
```
Amasus: This module contains the batching and retry-on-failure protocols. Batching and retrys occur by source, not destination for two reasons: there are more sources than destinations and are census block centroids, non census block group centroid destinations are guaranteed to be valid addresses on roads. Each failed source is  is retried one origin/destination pair at a time. Furthermore, it drops unroutable rows, to avoid silent passing of these errors downstream. Finally, it reshapes the response to the desired long-form output.

python/utils/driving_distance_matrix.py
```diff
@@ -91,63 +90,32 @@ def matrix_response_to_long_df(source_names, dest_names, distances) -> pd.DataFr
...
-def _coerce_negative_distances_to_null(df: pd.DataFrame) -> pd.DataFrame:
-    '''Replace any negative ``distance_m`` with NaN, in place.
+def _reject_negative_distances(df: pd.DataFrame) -> None:
+    '''Raise if any ``distance_m`` is negative.
 
-    ORS signals "no route" with null, which becomes NaN and is recovered by the
-    snap/retry path. A negative distance is never a valid real-world value; if a
-    routing backend ever emits one, treat it as no-route so it flows through the
-    same recovery and never reaches the output CSV. ``0`` is left untouched.
+    A negative driving distance is physically impossible, so it can only
```
Amasus: Guards against a negative driving distance appearing in the data matrix, which is a documented feature of some ORS versions. A "no route" response from ORS, represented as null (NaN), and a ``0`` distance are both kept.

python/utils/driving_distance_matrix.py
```diff
@@ -268,92 +236,40 @@ def _retry_sources_individually(failed_sources: list[str],
...
+    '''Build a long-form driving-distance DataFrame for every routable source x dest pair.
 
     Strategy:
         1. Batch sources into ORS matrix calls capped at ``MATRIX_CELL_LIMIT``
            cells (sources x dests).
         2. On batch failure (``OrsMatrixError``), defer the entire batch to
            per-source single-pair retries via the directions endpoint.
-        3. Origins ORS cannot route at all are snapped to their nearest
-           in-range haversine neighbor via ``estimate_origin``.
+        3. Pairs ORS still cannot route are dropped from the result — never
```
Amasus: 3. A pair ORS cannot route is dropped. This function does not raise or fail on a drop — the CLI compares requested source_ids against returned id_orig values and fails on any gap.

python/utils/driving_distance_matrix.py
```diff
@@ -383,22 +299,28 @@ def build_distance_matrix(*,
...
+    return df.dropna(subset=[DISTANCE_DISTANCE_M]).reset_index(drop=True)
 
 
 def resume_from_partial_output(output_path, source_ids, dest_ids):
-    '''Return ``(existing_df, remaining_pairs)`` for resuming a partial run.
-
-    Fixes the upstream geolib bug where an existing partial output caused the
-    CLI to return without writing anything. Instead, treat the existing file
-    as a warm-start cache and compute the pairs still to fetch.
+    '''Return ``(existing_df, remaining_pairs)`` for resuming from a previous run's partial output CSV.
```
Amasus: Problem with function name: try
identify_unmatched_pairs

Also, this function could be used to deal with updating potential_locations files (like what happened in Tarrant County). One of the root causes of #347. Probably worth a separate ticket.

python/scripts/generate_driving_distances_cli.py
Amasus, line 61: location-config? What is that? do you just mean the config?

Amasus: This code treats the Latitude/ Longitude as a first class data structure, when it is actually deprecated. I'm replacing it's need in derive_origins_and_destinations

python/scripts/generate_driving_distances_cli.py
```diff
@@ -205,6 +220,79 @@ def _assert_ors_reachable(matrix_url: str) -> None:
...
+    Args:
+        origins: Origin ids to list.
+        locations: Mapping from id to ``[longitude, latitude]``.
+
+    Returns:
+        A list of formatted message lines.
+    '''
+    lines = []
+    for origin in sorted(origins):
+        longitude, latitude = locations.get(origin, (None, None))
```
Amasus: (None, None) -> [None, None]

to match types.

python/scripts/generate_driving_distances_cli.py
Amasus: changing the code here for speed

Amasus: explicitly removing. I'm not seeing the value of having this call:

From Claude:
I tried to construct a case for keeping it and couldn't make one hold up. Three candidates, each falls apart:

"No side effects, safe to explore with" — but the side effect of a normal run is a partial CSV of real, successfully-routed data. That's not something to avoid; it's the whole point of this PR (never lose completed work). If you decide the county isn't viable, deleting a gitignored CSV is trivial. There's no harm being avoided.
"Cheaper way to just check" — the opposite is true. Probe mode always queries ORS for the entire county, every single time you run it, because it never touches identify_unmatched_pairs — no incremental resume, ever. Run it twice and you pay full cost twice. A normal run only pays full cost once; every rerun after that only re-queries what's still missing. Probe mode is strictly more expensive on any repeat use, not less.
"Explicitly signals intent (I'm just checking, not committing)" — same detection logic, same messages, same exit code either way. Reading "wrote N rows to <path>" vs. "no output written" communicates the same thing a flag name would.
So: it costs more, it protects against nothing, and it doesn't tell you anything a normal run wouldn't. I don't see what's missing — this reads like the flag exists because #325's original review kept it without deciding to, not because it does something the normal path can't.

Amasus (changes requested):

Amasus:
Opened #351 against this branch — a comment/docstring accuracy pass plus a few small code fixes (see the tables in the PR description for exactly what changed and why). #351 is a superset of the changes indicated in the comments on this PR.

@Abd1tus I'll approve #337 once you've approved #351.

python/scripts/generate_driving_distances_cli.py
abd1tus: It should have been a positional arg.  It's not a hard change, but has many touch points.  I can get it as part of this PR, or a followup (to reduce the size). What's your preference?

python/utils/ors_client.py
abd1tus: Done. The module docstring now says what ORS is, that we run it locally on a state extract, and what the matrix and directions endpoints each do for us. Commit 1e244999.

python/utils/driving_distance_matrix.py
abd1tus: Renamed in #351, which is merged into this branch already. New ticket for the locations file case: #359

python/scripts/generate_driving_distances_cli.py
abd1tus: Done in #351.

python/scripts/generate_driving_distances_cli.py
abd1tus: Done in #351.

abd1tus: Removed as part of #351.

Amasus: Lets leave it for now and just create a ticket for later. It's a nice to have clarifying comment.

python/utils/driving_distance_matrix.py
```diff
@@ -97,9 +97,10 @@ def matrix_response_to_long_df(source_names, dest_names, distances) -> pd.DataFr
 def _reject_negative_distances(df: pd.DataFrame) -> None:
     '''Raise if any ``distance_m`` is negative.
 
-    Guards against a negative driving distance appearing in the data matrix, 
-    which is a documented feature of some ORS versions. A "no route" response 
-    from ORS, represented as null (NaN), and a 0 distance are both kept.
+    A negative driving distance is never valid: ORS returns null for an unroutable
```
Amasus: This is great. Can you just get rid of the ticket numbers?

Amasus (changes requested):

python/utils/driving_distance_matrix.py
abd1tus: Done.

python/scripts/generate_driving_distances_cli.py
abd1tus: Ticket #360 created for this.

Amasus (approved):

Issue #350, opened 2026-09-01
generate_driving_distances_cli: replace free-text --state override with a --testing boolean


Amasus:
## Problem

\`--state <slug>\` lets an operator override state derivation with an arbitrary Geofabrik slug. Its only real purpose is handling the \`testing\` fixture, whose \`location\` field (\`testing\`) has no \`_<ST>\` suffix to derive a state from. But because it accepts any valid slug, nothing stops a real county config from being pointed at the wrong state entirely — e.g. a hypothetical \`Cook_County_GA\` config run with \`--state illinois\` would silently build/query the wrong ORS graph.

The "operators may legitimately point at a different state's graph" justification for keeping this general was checked directly this session and doesn't hold up: the buffered OSM extract already includes a 50km buffer around the *correct* state, which covers legitimate cross-border routing (e.g. Monongalia County WV routing into VA) without needing an override. Pointing at an unrelated state just breaks routing wholesale.

`python/tests/generate_driving_distances_cli_test.py::TestConfigDerivedState::test_explicit_state_flag_overrides_config_derivation` currently encodes this exact disproven justification in its own docstring ("operators may legitimately point at a different graph") — it exists specifically to protect the behavior this issue proposes removing.

## Proposed fix

Replace \`--state <slug>\` with a boolean flag (e.g. \`--testing\`). When set, hardcode \`state = 'georgia'\` (the testing fixture's actual state) instead of accepting an arbitrary slug. When unset, state derivation from \`config.location\`'s trailing \`_<ST>\` suffix works exactly as it does today.

This closes the wrong-state risk for real county configs while still covering the one legitimate use case (the \`testing\` fixture).

## Scope (single CLI script + its tests + one doc line)

- \`python/scripts/generate_driving_distances_cli.py\`: swap the \`--state\` argument for the boolean flag in \`build_arg_parser\`; replace the \`args.state\`-based branching in \`main()\`; update the error message in the location-derivation-failure branch (currently suggests \`--state georgia\`, needs to suggest the new flag). \`_reject_unknown_slug\` stays, applied only to the derived-from-location path.
- \`docs/to_run.md\`: update the \`--state\` example/mention.
- \`python/tests/generate_driving_distances_cli_test.py\`:
  - Delete \`test_rejects_unknown_state_slug\` (no more free-text value to reject).
  - Rename/adjust \`test_state_flag_no_longer_required_on_cli\` for the new flag name.
[...]

PR #351, opened 2026-09-01
Make #325 branch comments readable: docstring/comment accuracy pass + minor code fixes

### LINKED ISSUE #325: generate_driving_distances_cli: dropped/unrouted origins are silent, undocumented, and validated too late

Amasus:
## Summary

This is primarily a comment/docstring readability pass on top of `fix/325-unrouted-origin-handling` (PR #337) — going line by line through the files that branch touches, verifying every claim a comment or docstring makes against the actual code/git/issue history, and fixing what didn't hold up. Also removes `--check-bad-locations` entirely, resolving #325's own deferred "keep vs remove" question.

**Everything in this diff beyond the two tables below is a comment, docstring, or test-docstring change** — wording, accuracy, or removing stale/unverified claims. No behavior changes hide in that part of the diff.

## Scope note

This branch touches slightly more than #325/PR #337's original scope. While verifying `filter_distance_data`'s docstring accuracy, we found and fixed an unused `log` parameter (dead since the function's original commit, unrelated to #325) — that pulled in `python/solver/model_data.py`, `model_run.py`, `conftest.py`, and `model_data_test.py`, none of which #325 or PR #337 touch. Flagging this explicitly since it's outside the original ticket's boundary, even though it's a small, low-risk change.

## Code changes (behavior, not wording) — table as of this PR

| File | Change | Reason |
|---|---|---|
| `driving_distance_matrix.py` | `get_missing_origins` → `get_origins_with_blank_distances` (rename) | Old name didn't match behavior; new name matches existing "blank distance_m" terminology already used elsewhere in the file |
| `driving_distance_matrix.py` | `resume_from_partial_output` → `identify_unmatched_pairs` (rename, + import/call-site update in `generate_driving_distances_cli.py`) | Old name implied an action ("resume") the function doesn't perform — no I/O beyond a read, no querying, no writing |
| `generate_driving_distances_cli.py` | `--check-bad-locations` CLI flag removed from `build_arg_parser` | Resolves #325's own deferred "keep vs remove" question: no benefit over a normal run, strictly more expensive (always full-county query, never benefits from resume) |
| `generate_driving_distances_cli.py` | `main()`'s `if args.check_bad_locations:` probe-mode branch removed entirely | Direct consequence of removing the flag |
| `generate_driving_distances_cli.py` | `_report_unrouted_origins`'s `output_path` parameter changed from optional (`str \| None = None`) to required (`str`); the `if output_path is None` branch collapsed | Direct consequence of removing check-bad-locations — probe mode was the only caller with no output path |
| `generate_driving_distances_cli.py` | `derive_origins_and_destinations`: origin-extraction loop vectorized (`.iterrows()` over `blocks_gdf` → column ops) | `.iterrows()` is a known-slow anti-pattern; no polymorphism need for this specific loop |
| `generate_driving_distances_cli.py` | `derive_origins_and_destinations`: destination-extraction loop vectorized (`.iterrows()` over `pots_df` → `.str.split(',', expand=True).astype(float)` on the whole `Lat, Lon` column) | Same as the origins loop — `.iterrows()` anti-pattern, no per-row need once the schema was reduced to one combined-column case |
| `generate_driving_distances_cli.py` | `derive_origins_and_destinations`: `_pick_coord_extractor` deleted; separate `Latitude`/`Longitude`-column support removed, destinations parsed only via combined `Lat, Lon` | Verified the two-schema premise was false — no real committed data ever required separate columns |
| `generate_driving_distances_cli.py` | `derive_origins_and_destinations`: malformed-coordinate validation added (comma-count check on `Lat, Lon`) | Re-implements, in vectorized form, validation that existed in the deleted `_pick_coord_extractor.extract_combined` (which raised on `len(parts) != 2`) — not new protection, carried over |
| `generate_driving_distances_cli.py` | `derive_origins_and_destinations`: the explicit "neither schema present" `ValueError` is gone — a missing `Lat, Lon` column now raises a raw `KeyError` instead | Considered and decided that adding error handling back would mean treating this one field differently from every other column read in this function — chose consistency over extra error handling |
| `generate_driving_distances_cli.py` | `_origin_lines`'s default value: `locations.get(origin, (None, None))` → `locations.get(origin, [None, None])` (tuple→list) | No behavioral change, just cleanup to keep types consistent |
| `model_data.py` / `model_run.py` | `filter_distance_data`'s `log` parameter removed from the signature; call sites updated | Verified via git history it's been unused since the function's original commit (Dec 2025) — dead parameter, not a recent regression |

## Test reorganization — `generate_driving_distances_cli_test.py`

**Before:**

| Class | Tests |
|---|---|
| `TestArgParser` | `test_requires_location_config`, `test_parses_location_config`, `test_default_logdir`, `test_state_defaults_to_none`, `test_explicit_state_flag`, `test_server_and_logdir_overrides` |
| `TestWriteOutputCsv` | `test_writes_only_three_columns_in_order` |
| `TestDeriveOriginsAndDestinations` | `test_handles_combined_lat_lon_column`, `test_raises_when_no_coord_columns` |
| `TestMain` | `test_writes_csv_at_expected_path`, `test_write_path_exits_nonzero_but_still_writes_partial_csv`, `test_resume_skips_pairs_already_in_output`, `test_resume_short_circuits_when_nothing_remains`, `test_short_circuit_exits_nonzero_on_blank_distance_row` |
| `TestStateValidation` | `test_rejects_unknown_state_slug`, `test_state_flag_no_longer_required_on_cli` |
| `TestConfigDerivedState` | `test_derives_state_from_config_location_when_no_flag`, `test_explicit_state_flag_overrides_config_derivation`, `test_errors_when_neither_flag_nor_derivable_location` |
| `TestInContainerHealthCheck` | `test_exits_when_ors_unreachable` |

**After:**

| Class | Tests | Notes |
|---|---|---|
| `TestArgParser` / `TestWriteOutputCsv` / `TestDeriveOriginsAndDestinations` / `TestInContainerHealthCheck` | unchanged | |
| `TestMain` | `test_writes_csv_at_expected_path` | Narrowed to match its own "smoke-test the happy path" docstring |
| `TestResumeBehavior` *(new)* | `test_resume_skips_pairs_already_in_output`, `test_resume_short_circuits_when_nothing_remains` | Pure resume/partial-fetch mechanics |
| `TestUnroutedOriginReporting` *(new)* | `test_write_path_exits_nonzero_but_still_writes_partial_csv`, `test_short_circuit_exits_nonzero_on_blank_distance_row` | Both verify `_report_unrouted_origins` firing + exit 1 + partial data kept — grouped by what's verified, not which branch triggers it |
| `TestStateResolution` *(merged)* | `test_derives_state_from_config_location_when_no_flag`, `test_errors_when_neither_flag_nor_derivable_location`, `test_rejects_unknown_state_slug`, `test_explicit_state_flag_overrides_config_derivation` | Merges `TestStateValidation` + `TestConfigDerivedState`. Dropped `test_state_flag_no_longer_required_on_cli` (verified exact duplicate of `TestArgParser::test_state_defaults_to_none`). Kept the override test — its deletion belongs with #350's implementation, not this PR. |

## Follow-ups filed, not in this PR

- **#350** — replace `--state <slug>`'s free-text override with a `--testing` boolean. The "operators may legitimately point at a different state's graph" justification (encoded in `test_explicit_state_flag_overrides_config_derivation`, kept in this PR) was checked directly and doesn't hold up.
- The two dead `--check-bad-locations` tests are already removed as part of this PR; a stale dual-schema test (`test_returns_locations_dict_and_id_lists`) was also removed after confirming its unique coverage (source-side coordinate extraction) was folded into `test_handles_combined_lat_lon_column` first.
- One known open item: a missing `Lat, Lon` column now raises a raw `KeyError` instead of a clear message (see code-change table above) — deliberate per the reason given, flagging in case reviewers disagree with that call.

### DIFF
--- a/docs/to_run.md
+++ b/docs/to_run.md
@@ -113,7 +113,7 @@ Both Gwinnett_G**A**_configs/Gwinnett* and Gwinnett_G**a**_configs/Gwinnett* wil
 
 ## Generating driving distances
 
-The solver consumes driving-distance CSVs at `datasets/driving/<Loc>_<ST>/<Loc>_<ST>_driving_distances.csv`. The `generate_driving_distances_cli` script builds those CSVs from existing project data (TIGER block centroids + the `<Loc>_<ST>_potential_locations.csv` already used by the solver) by routing every origin × destination pair through a locally-hosted OpenRouteService (ORS) container.
+The solver consumes driving-distance CSVs at `datasets/driving/<Loc>_<ST>/<Loc>_<ST>_driving_distances.csv`. The `generate_driving_distances_cli` script builds those CSVs from existing project data (TIGER block centroids + the `<Loc>_<ST>_potential_locations.csv` already used by the solver) by routing every origin × destination pair through a locally-hosted OpenRouteService (ORS) container. **USAGE NOTE:** This CLI is designed to only take local config, `<Loc>_<ST>_potential_locations` and census files, not to read from the database or make census pulls. See [input_files.md](input_files.md)for more information on generating or obtaining these files.
 
 ```bash
 python3 run.py generate_driving_distances_cli \
@@ -136,16 +136,8 @@ State slugs are full Geofabrik names (`georgia`, `new-york`, `district-of-columb
 
 ### Unroutable origins fail the run
 
-Every populated census block needs a real driving distance for the solver to run, so an origin ORS cannot route is an error, not a warning. When any origin is unrouted, the CLI still writes the CSV with everything that was successfully computed (completed work is never lost), then **exits non-zero** and prints each missing origin with its id and lat/lon. Fix the underlying data (typically a bad block centroid) and rerun — the resume logic reads the existing CSV and fetches only the missing pairs. Rows patched into the CSV by hand are likewise recognized as satisfied on the next run.
+Every populated census block needs a real driving distance for the solver to run, so an origin ORS cannot route is an error, not a warning. When any origin is unrouted, the CLI still writes the CSV with everything that was successfully computed (completed work is never lost), then **exits non-zero** and prints each missing origin with its id and lat/lon. Fix the underlying data (typically a bad block centroid) and rerun — the resume logic reads the existing CSV and fetches only the missing pairs. Rows patched into the CSV by hand are likewise recognized as satisfied on the next run. Note the check is origin-level: an origin that routes to some destinations but not others is not flagged here; that gap is caught downstream at model-run time. TODO: update after #338 is implemented.
 
-### Pre-flight probe: `--check-bad-locations`
-
-```bash
-python3 run.py generate_driving_distances_cli --check-bad-locations \
-  -l datasets/configs/<config_set>/<config>.yaml
-```
-
-Probe-only mode: performs the same ORS routing but writes **no** output CSV, then reports unroutable origins and exits non-zero if it found any (0 when all origins routed). Useful as a pre-flight check when onboarding a new county — it surfaces bad centroids before you commit to treating a generated matrix as final. Note the check is origin-level: an origin that routes to some destinations but not others is not flagged here; that gap is caught downstream at model-run time.
 
 ### Manual lifecycle (debugging / repeated experiments)
 
--- a/docs/to_run.md
+++ b/docs/to_run.md
@@ -113,7 +113,7 @@ Both Gwinnett_G**A**_configs/Gwinnett* and Gwinnett_G**a**_configs/Gwinnett* wil
 
 ## Generating driving distances
 
-The solver consumes driving-distance CSVs at `datasets/driving/<Loc>_<ST>/<Loc>_<ST>_driving_distances.csv`. The `generate_driving_distances_cli` script builds those CSVs from existing project data (TIGER block centroids + the `<Loc>_<ST>_potential_locations.csv` already used by the solver) by routing every origin × destination pair through a locally-hosted OpenRouteService (ORS) container. **USAGE NOTE:** This CLI is designed to only take local config, `<Loc>_<ST>_potential_locations` and census files, not to read from the database or make census pulls. See [input_files.md](input_files.md)for more information on generating or obtaining these files.
+The solver consumes driving-distance CSVs at `datasets/driving/<Loc>_<ST>/<Loc>_<ST>_driving_distances.csv`. The `generate_driving_distances_cli` script builds those CSVs from existing project data (TIGER block centroids + the `<Loc>_<ST>_potential_locations.csv` already used by the solver) by routing every origin × destination pair through a locally-hosted OpenRouteService (ORS) container. **USAGE NOTE:** This CLI is designed to only take local config, `<Loc>_<ST>_potential_locations` and census files, not to read from the database or make census pulls. See [input_files.md](input_files.md) for more information on generating or obtaining these files.
 
 ```bash
 python3 run.py generate_driving_distances_cli \
--- a/python/CLAUDE.md
+++ b/python/CLAUDE.md
@@ -32,6 +32,8 @@ When code is changed or refactored, update any affected docstrings to stay accur
 
 Inline comments should explain *why*, not restate *what* the code does. Keep them concise and follow PEP 8 (single space after `#`, sentence case).
 
+One exception: a short **signpost** — a noun/verb phrase, not a sentence — may mark the start of a multi-line phase in a function long enough that skimming it cold is hard. Signposts mark phases, not statements: if the block is a single self-documenting call (e.g. `_reject_negative_distances(df)`), a signpost just restates the function name and should be deleted, not shortened. One per phase. Rationale still belongs at the phase's own definition — a function's docstring — never repeated at the call site.
+
 ## Test-Driven Development
 
 All code must be written using TDD:
--- a/python/CLAUDE.md
+++ b/python/CLAUDE.md
@@ -34,6 +34,10 @@ Inline comments should explain *why*, not restate *what* the code does. Keep the
 
 One exception: a short **signpost** — a noun/verb phrase, not a sentence — may mark the start of a multi-line phase in a function long enough that skimming it cold is hard. Signposts mark phases, not statements: if the block is a single self-documenting call (e.g. `_reject_negative_distances(df)`), a signpost just restates the function name and should be deleted, not shortened. One per phase. Rationale still belongs at the phase's own definition — a function's docstring — never repeated at the call site.
 
+A comment can be specific and falsifiable and still misattribute *why* — e.g. attributing a rejection rule to the specific algorithm consuming the data, when the rule is actually a general data-integrity constraint the algorithm has nothing to do with. Check that the named mechanism is the right one, not just that *some* mechanism is named.
+
+A test's docstring should describe what the test verifies, not re-derive why the underlying code behaves that way — that reasoning belongs at the production code's own definition, not the test.
+
 ## Test-Driven Development
 
 All code must be written using TDD:
@@ -43,6 +47,10 @@ All code must be written using TDD:
 3. **Refactor** — clean up while keeping all tests green.
 
 When modifying existing code, update the associated tests before or alongside the change, and confirm all relevant tests pass before considering the work done.
+
+When deleting or consolidating tests, verify every assertion is covered elsewhere first — not just the ones tied to your reason for removing it. Also check for exact duplicates hiding under different names when merging test classes.
+
+A test whose entire premise guards a since-deleted code path has no value even if its assertions still pass — delete it rather than patching its literal values to keep it green.
 
 ## Key Conventions
 
--- a/python/scripts/generate_driving_distances_cli.py
+++ b/python/scripts/generate_driving_distances_cli.py
@@ -19,6 +19,8 @@ from python.solver.constants import (
     DISTANCE_DISTANCE_M,
     DISTANCE_ID_DEST,
     DISTANCE_ID_ORIG,
+    POT_LOC_LAT_LON,
+    POT_LOC_LOCATION,
     TIGER20_GEOID20,
     TIGER20_INTPTLAT20,
     TIGER20_INTPTLON20,
@@ -113,12 +115,8 @@ def derive_origins_and_destinations(config):
     locations = dict(zip(source_ids, coords))
 
     # get destination id and location coordinates from potential_location data
-    destination_ids = potential_locations_df['Location'].tolist()
-    malformed_coordinates = potential_locations_df['Lat, Lon'].str.count(',') != 1
-    if malformed_coordinates.any():
-        bad = potential_locations_df.loc[malformed_coordinates, 'Lat, Lon'].tolist()
-        raise ValueError(f'Lat, Lon column: expected exactly one comma per value, got: {bad}')
-    joint_coords = potential_locations_df['Lat, Lon'].str.split(',', expand=True).astype(float)
+    destination_ids = potential_locations_df[POT_LOC_LOCATION].tolist()
+    joint_coords = potential_locations_df[POT_LOC_LAT_LON].str.split(',', expand=True).astype(float)
     dest_coords = joint_coords[[1, 0]].values.tolist()  # column 0 is lat, column 1 is lon — reorder to [lon, lat]
     locations.update(dict(zip(destination_ids, dest_coords)))
 
--- a/python/scripts/generate_driving_distances_cli.py
+++ b/python/scripts/generate_driving_distances_cli.py
@@ -1,9 +1,9 @@
 '''CLI: generate the driving-distance CSV for a given county config.
 
-Reads census block centroids and the potential-locations CSV the solver
-consumes, calls OpenRouteService to build a driving-distance matrix, and
-writes the canonical three-column long-form CSV to
-``datasets/driving/<location>/<location>_driving_distances.csv``.
+Reads census block centroids and the potential-locations CSV from disk
+(no DB read or census pulls enabled), calls OpenRouteService to build a 
+driving-distance matrix, and writes the canonical three-column long-form 
+CSV to ``datasets/driving/<location>/<location>_driving_distances.csv``.
 '''
 import argparse
 import os
@@ -28,8 +28,8 @@ from python.solver.model_data import get_blocks_gdf, load_potential_locations_cs
 from python.utils.directory_constants import DRIVING_DIR
 from python.utils.driving_distance_matrix import (
     build_distance_matrix,
-    get_missing_origins,
-    resume_from_partial_output,
+    get_origins_with_blank_distances,
+    identify_unmatched_pairs,
 )
 from python.utils.ors_setup import GEOFABRIK_STATE_SLUGS, state_slug_from_location
 from python.utils.ors_url import resolve_ors_url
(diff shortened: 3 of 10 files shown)


abd1tus:
Good changes.  The renames are real improvements, the `--check-bad-locations`
removal is clean (docs updated too), and we checked the schema change independently: both
committed location files (`testing` and `Tarrant_County_TX`) carry a `"Lat, Lon"` column,
and in the testing fixture it agrees with the separate `Latitude`/`Longitude` columns on
every row, so dropping the separate-column path changes nothing for existing data. The
vectorized destination parsing gives identical ids and coordinates to the old loop on the
real Tarrant file (368 rows) and on whitespace variants. Happy to merge with one fix:

**One test fails on this branch:** `test_raises_when_no_coord_columns`
(`generate_driving_distances_cli_test.py`, line 100). It checks that a locations file
with no `"Lat, Lon"` column produces a clear error — but with the old format-detection
code deleted, a missing column now blows up inside pandas as a bare `KeyError: 'Lat, Lon'`
(at `generate_driving_distances_cli.py:117`), so the expected `ValueError` never appears.
Fresh run of the two test files: base branch 47 passed; this branch 42 passed, 1 failed.

**Potential fix, two small parts:**

1. Before using the column, check it exists and raise a clear error — something like:

       if 'Lat, Lon' not in potential_locations_df.columns:
           raise ValueError(f"potential locations file has no 'Lat, Lon' column "
                            f"(columns found: {list(potential_locations_df.columns)})")

2. The test still matches on the old two-schema wording (`match='Latitude/Longitude'`),
   so update its expected text to `'Lat, Lon'` and its docstring to match. Worth keeping
   rather than deleting under the "guards a deleted code path" rule: its real purpose (a
   bad header gets a clear error) survives the schema change.

That also answers the judgment call you flagged: we'd lean toward keeping the clear error.
A bad header is now *more* likely, not less — the old detection accepted spellings like
`"Lat, Long"` (which is literally what this column was called until it was renamed in
8a0e865e), and this column already gets special treatment on the very next line (the
comma-count check).

Three smaller things, none blocking:

- `_reject_negative_distances`'s new docstring says a negative distance "is a documented
  feature of some ORS versions" — we couldn't source that, and it contradicts both the
  function's own error message ("never a valid driving distance") and our earlier verified
  finding (ORS v8.0.0 returns null, not a negative, for no-route cells). Suggest restoring
  the previous sentence.
- The old loop turned every destination name into text explicitly; the new vectorized
  version doesn't. If a location file ever has numeric-looking names (a precinct called
  "101"), they'd load as numbers and then fail to match their text ids elsewhere — the
  same numeric-vs-text id mismatch behind #305 on the origins side. It happens to be safe
  today because `load_potential_locations_csv` forces the `Location` column to text one
  layer down, but the origins line kept its `.astype(str)` and destinations should match
  it. One-line fix.
- `POT_LOC_LOCATION` / `POT_LOC_LAT_LON` exist in `solver/constants.py` — the new code
  hardcodes both literals (same goes for the snippet above).

Amasus:
## Summary of changes made and response to review

Thanks for the thorough review. Here is a summary of your findings and our calls on each.

| Concern raised | Discussed before this review? | What we decided, and why |
|---|---|---|
| Renames (`get_missing_origins`→`get_origins_with_blank_distances`, `resume_from_partial_output`→`identify_unmatched_pairs`) are real improvements | Yes | Made deliberately — old names didn't match actual behavior |
| `--check-bad-locations` removal is clean | Yes | Deliberate removal — resolved #325's own deferred question, no benefit over a normal run |
| Dropping separate `Latitude`/`Longitude` column support is safe | Yes | Verified the two-schema premise was false before removing it — no real committed data ever required separate columns |
| Vectorized destination parsing matches the old loop's output | Yes | Verified via the vectorization logic itself |
| `test_raises_when_no_coord_columns` fails — missing `Lat, Lon` now raises bare `KeyError`, not `ValueError` | Yes | **Keeping the `KeyError`.** Dropping the test entirely. See below — reconsidered this morning with the codebase checked end to end, not just this file |
| `_reject_negative_distances`'s "documented feature of some ORS versions" is unsourced and contradicts the function's own message | Yes | **Deliberate** — left in as review-comment bait to see whether the claim would get challenged or accepted. See #225 below |
| Destination names not coerced to text (`.astype(str)`), same numeric-vs-text risk as #305 | No, new finding | Belongs with #305 — same bug class, tracked centrally there rather than patched ad hoc here |
| Hardcoded `'Lat, Lon'`/`'Location'` instead of `POT_LOC_LAT_LON`/`POT_LOC_LOCATION` constants | Yes | Fixed — both the production code and its tests now use the constants |

## On the missing-column check

We rechecked this by looking at every other place this exact column gets read, not just
this one function: `load_potential_locations_csv` (model_data.py), `db_import_potential_locations_cli.py`,
and the `build_distance_data` pipeline. None of them validate for a missing `Lat, Lon`
column either — they all use the `POT_LOC_LAT_LON` constant and let a missing column
fail naturally. Adding a bespoke check only in this file would make it the outlier, not
bring it in line with an existing convention. `Lat, Lon` is also populated by copy-paste
from a Google Maps output (see `docs/input_files.md`) — a stable intake format with zero
observed anomalies in any committed file, checked across every `*_potential_locations.csv`
in the repo.

Resolution: keeping the `KeyError` as-is. We also removed the malformed-comma-count
check that was previously added on this same column, for the same reason — and deleted
`test_raises_when_no_coord_columns` rather than patch its `match=` text, since there's no
`ValueError` left for it to verify.

If the underlying concern is real — that input data sources deserve column-presence
validation somewhere — it's a broader question than this one column. A survey of every
input data source (potential_locations, config fields, driving-distances CSVs) and a
single, consistent policy for whether/where column-presence gets validated would be a
more useful next step than a one-off check added wherever a reviewer happens to be
looking.

## On the `_reject_negative_distances` docstring

Traced this back before responding. #225 (filed by Chad, 2026-06-01) already
investigated and settled this exact question:

> "we confirmed empirically that ORS **v8.0.0** `/v2/matrix/driving-car` with
> `metrics:["distance"]` returns JSON **null** (not `-1`) for an unroutable cell — so
> there is no live source of negative distances today. **The original `-1` concern
> (raised by web research) was refuted against the live georgia graph.**"

That ticket is also the origin of the negative-distance guards this function
implements. We can discuss how this should be documented, human to human, to convey the right information to future readers. A word of advice to the bot: since the answer was already on record in #225, it's worth
checking open tickets/history for context like this before flagging a claim as
unsourced.

PR #352, opened 2026-09-03
fix(scripts): replace --state override with --testing boolean (#350)

Amasus:
## Summary
- `--state <slug>` accepted any Geofabrik slug with no relation to the config's actual location, so a mistyped or misconfigured county could silently build/query the wrong ORS graph (e.g. `Cook_County_GA` run with `--state illinois`).
- Replaces it with a `--testing` boolean that hardcodes the testing fixture's known state (`georgia`). Real configs always derive state from their own `location` field — no override path remains.
- `run.py`'s orchestrator had the same free-text `--state` override one layer up, plus duplicate derivation logic. It's updated to mirror the same fix: `--testing` only, and it now forwards `--testing` alone to the in-container script rather than a resolved state value — the container re-derives from the same `-l` config instead of trusting a forwarded value.
- `_reject_unknown_slug` is removed from the CLI script: it validated against `GEOFABRIK_STATE_SLUGS`, but `state_slug_from_location` already raises for any invalid slug, so the extra check was unreachable dead code once the free-text override was gone.
- Fixed an ordering bug found in passing: `main()` loaded the config before checking ORS reachability, which broke the reachability test's premise (config load could fail first, masking whether the reachability check itself worked).

## Files changed
docs/to_run.md
python/scripts/generate_driving_distances_cli.py
python/tests/generate_driving_distances_cli_test.py
python/tests/run_generate_driving_orchestration_test.py
python/tests/run_py_test.py
run.py

Note that run.py and surrounding tests were out of scope for the original ticket, but discovered to be necessary during the run. Similarly was changing the test fixtures. But I was here, and I didn't want to let bad coding practices slide.

Also note that the following decisions were intentional and carefully thought through and should not be overturned by a machine without a human to human conversation:
* _ = state_slug_from_location(config.location) — result thrown away. Looks like dead/leftover code. Real reason: state isn't used for anything else in this script — all state-consequential work (extract build, ORS boot) happens in run.py. Someone could "clean this up" by deleting the call entirely, silently losing the typo check.
* The try/except ValueError around it — looks redundant since state_slug_from_location already raises a clear message on its own. Real reason: it's not about the message, it's about the exit code. Unwrapped, a bad location exits with Python's default code 1, colliding with this script's own "unrouted origins" meaning for exit 1. Someone could "simplify" this away and reintroduce that collision.
* _reject_unknown_slug deleted entirely — a reviewer sees removed validation and assumes it was accidental or unsafe. Real reason: it's mathematically dead code once --state free text is gone — state_slug_from_location already validates internally, and 'georgia' is a hardcoded literal. Someone could "restore" it as a safety net.
* Reachability check moved before config load — looks like an arbitrary reorder. Real reason: preserves test_exits_when_ors_unreachable's premise (distinguishing "ORS down" from "bad config path"). Someone could reorder back "for readability."
* The run.py forward---testing-only decision is already self-documented in a code comment, so hopefully that one's safe.

### DIFF
--- a/python/tests/run_py_test.py
+++ b/python/tests/run_py_test.py
@@ -225,7 +225,8 @@ class TestGenerateDrivingDistancesOrchestration:
     @patch('run.run_command')
     def test_errors_before_docker_when_state_not_derivable(
             self, mock_run_command, mock_subprocess_run, unused_mock_healthy, tmp_path):
-        '''No --state, location='testing' -> exit 2 before any docker activity.'''
+        '''No --testing, location='testing'. Therefore no state can be read from location 
+        and none supplied. -> exit 2 before any docker activity.'''
         del unused_mock_healthy, mock_run_command
         cfg = tmp_path / 'cfg.yaml'
         cfg.write_text('config_set: testing\nlocation: testing\n')
--- a/python/tests/run_py_test.py
+++ b/python/tests/run_py_test.py
@@ -47,7 +47,7 @@ class TestGenerateDrivingDistancesOrchestration:
         '''Default path: bring ORS up, run the matrix, then bring ORS down.'''
         del unused_mock_healthy
         with patch('sys.argv', ['run.py', 'generate_driving_distances_cli',
-                                '--state', 'georgia', '-l', 'cfg.yaml']):
+                                '--testing', '-l', 'cfg.yaml']):
             run_module.main()
         ors_up_calls = [
             c for c in mock_subprocess_run.call_args_list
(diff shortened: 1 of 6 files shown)


run.py
abd1tus: Found an outdated --state mention here.

python/tests/run_py_test.py
abd1tus: Another --state mention.

abd1tus (changes requested):

Amasus:
1. Thank you for giving me human comments instead of machine comments. It seems dumb, but it's a huge quality of life improvement for me. I appreciate that extra effort on your part. Immensely.
2. Updated the comments. thank you for catching. I should have done a grep instead of relying on claude.
3. I just want to say how nice and easy it was to make this change. A sign of good design when a flag can be swapped out so smoothly. I appreciate you.

run.py
```diff
@@ -400,11 +399,8 @@ def main():
                 state = state_slug_from_location(location)
             except ValueError as exc:
                 print(
-                    f"Couldn't derive state from {config_path}: {exc}.\n"
-                    f"Either rename the location to end in _<ST> "
-                    f"or pass an explicit override:\n"
-                    f"  python3 run.py generate_driving_distances_cli "
-                    f"--state georgia -l {config_path}",
+                    f"Couldn't derive state from config (location={location}; {exc}).\n"
```
abd1tus, line 402: Minor bug here.  "location" is being used out of scope.

PR #364, opened 2026-09-21
Tarrant County 2026 delivery analysis

Amasus:
Tarrant County 2026 delivery analysis: the config, regressions, and result tables backing the 2026 polling location findings.

This branch was cut directly from `main` rather than flowing through `dev`, deliberately — the in-progress engineering work on `dev` is unrelated to this delivery and the analysis needs to ship without waiting on it. It therefore targets `main`.

## What's here

- **New Basic_analysis config** — `Tarrant_County_original_and_fair_2026.r`, covering the 2024/2025/2026-actual/2026-proposed original configs plus the 215-open fair config.
- **Closure analysis** (`Tarrant_County_TX_closure_analysis.r`) — Latine/White interaction regressions (LPM and logit) against poll closures, and two result tables written to cloud storage:
  - `precinct_demographics_by_year.csv` — population assigned to each precinct by year, with the 2024→2026 change.
  - `block_distance_by_year.csv` — per census block, the assigned polling location and distance for each of 2024, 2025 and 2026, with demographic populations, percent Hispanic, and the 2024→2026 percent change in distance.
- **Supporting changes** to `Basic_analysis.r`, `graph_functions.R`, and the Tarrant potential-locations dataset.

## Notes for review

- `block_distance_by_year.csv` uses `distance_m` from the `results` table, which is the metres from a block to its assigned poll. Note that `weighted_dist` in that same table is `population x log(distance)` — computed before the log distances are exponentiated back — so it is not comparable to `weighted_dist` in the `precinct_distances`/`residence_distances` tables, which is `demo_pop x metres`. Only `distance_m` is used here.
- The demographic melt deliberately takes the six columns the `*_distances` tables use (`population`, `hispanic`, `white`, `black`, `native`, `asian`) rather than all ten available in `results`; the remaining four (`non_hispanic`, `pacific_islander`, `other`, `multiple_races`) would double-count people.
- `demo_pop` and `pct_hispanic` sit on the left of the `dcast` formula because census counts are fixed across the three years — only the assignment changes. Verified: all 23,191 origins are common to the 2024 and 2025 result files with zero differing population counts.
- `droplevels` on `descriptor` after the three-year filter is load-bearing. `descriptor` is a factor whose levels come from the whole `DESCRIPTOR_DICT_ORIG`, so without it `dcast` emits empty `id_dest_proposed` / `distance_m_proposed` columns.
- Precinct rows that previously fell through as `NA` are now labelled: `Closed` where the precinct is absent in 2026 (882 rows / 147 precincts), and `enacted_2025_only` where absent from both 2024 and 2026 (36 rows / 6 precincts).

## Verification

Latest run uploaded 15 objects to `gs://equitable-polling-analysis/equitable-polling-locations-analyses/Tarrant_County_TX_exploration/20260920-153332/`. `block_distance_by_year.csv` is 139,146 rows (23,191 blocks x 6 demographics); local and remote MD5s match.


antisocialscientist (approved):
Straightforward and looks good

